PR #45, opened 2023-03-12
36 integrate scheduler prototype into mvp

martinjonsson01:
I've split up the work of this PR into three sub-PRs to make it more digestible for reviewers.

See #36 for the sub-tasks. These sub-tasks will be merged into `36-integrate-scheduler-prototype-into-mvp` one at a time, and then this PR can be merged without review (as each commit it consists of will have been reviewed already). 

Closes #36.

PR #46, opened 2023-03-12
42 system ordering

### LINKED ISSUE #42: System ordering
## Reason
An explicit ordering of every type of system is necessary for a precedence graph to be generated.

## Requirements
- [x] There is an ordering of systems which distinguishes between:
  - [x] which systems need to run before/after others;
  - [x] which systems can run concurrently.

## Dependencies
None.

martinjonsson01:
Implements a partial ordering of systems, such that systems that can not run concurrently are ordered before or after each other. 

It does not implement `PartialOrd` on `System`, rather it creates a new trait `Orderable`. This is done because if `PartialOrd` is implemented on a type, it requires `PartialEq` to also be implemented and it has a list of things that need to be consistent with `PartialEq` (see [documentation](https://doc.rust-lang.org/core/cmp/trait.PartialOrd.html)). If those conditions were to be fulfilled, then equality for the `System` type would be based on whether systems can run concurrently, not on if the system instances are actually equal — which would have been very unintuitive.

Closes #42.


crates/scheduler/proptest-regressions/precedence.txt
```diff
@@ -0,0 +1,10 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
```
JacobBredin, line 6: ok, was wondering about this 👍

crates/ecs/src/lib.rs
```diff
@@ -275,15 +278,67 @@ pub trait System: Debug + Send + Sync {
...
-#[derive(Debug, Clone, Copy, Ord, PartialOrd, Eq, PartialEq)]
+#[derive(Debug, Clone, Copy, Ord, PartialOrd, Eq, PartialEq, Hash)]
 pub enum ComponentAccessDescriptor {
     /// Reads from component of provided type.
     Read(TypeId),
     /// Reads and writes from component of provided type.
     Write(TypeId),
 }
 
+impl ComponentAccessDescriptor {
```
JacobBredin, line 305: very nice things 👍

JacobBredin (approved):
LGTM, I have a question:

I agree with what you stated in the PR comment about it being unintuative to implement `PartialOrd` for `System`. However, I do not see why systems need to have an "inherent" order? Isn't the only thing that needs to be determined whether read and writes overlapp or not? 

I can imagine that creating a schedule is not an easy problem, so maybe I'm missing something. Like to hear your thoughts 😋

crates/scheduler/src/precedence.rs
```diff
@@ -0,0 +1,276 @@
...
+    prop_compose! {
+        fn arb_ordered_systems(max_system_count: usize)
+                              (systems_count in 1..=max_system_count)
+                              (systems in hash_set(arb_system(systems_count), systems_count))
+                              -> HashSet<Box<dyn System>> {
+            systems
+        }
+    }
+
+    /// NOTE: this assertion does _not_ hold for systems that mutually read and write to each others
```
EdvinNilsson, line 237: Good note! I was thinking about this scenario when reading the implementation and this explains it.

EdvinNilsson (approved):

PR #47, opened 2023-03-17
44 thread pool

### LINKED ISSUE #44: Thread pool
## Reason
A thread pool executor is necessary in order to be able to execute systems concurrently on multiple threads.

## Requirements
- [x] Systems execute on a thread pool according to a given schedule.
- [x] Threads in the thread pool are reused.
- [x] The thread pool can shut down gracefully.
- [x] Threads propagate panics to main thread.
- [x] There is work-stealing occurring to load-balance between workers.

## Dependencies
None.

martinjonsson01:
Implements a thread pool that executes systems on multiple worker threads, taking care to handle them correctly (i.e. graceful shutdown/worker panic).

Also prepares a bit for #43 (which is why you'll see a few to-dos that reference it).

Closes #44


crates/ecs/src/lib.rs
```diff
@@ -152,9 +215,15 @@ impl<'systems> Schedule<'systems> for Unordered<'systems> {
         Self(systems)
     }
 
-    fn next_batch(&mut self) -> Vec<&dyn System> {
-        self.0.iter().map(|system| system.as_ref()).collect()
+    fn currently_executable_systems(&mut self) -> Vec<SystemExecutionGuard<'systems>> {
+        self.0
+            .iter()
+            .map(|system| system.as_ref())
+            .map(|system| SystemExecutionGuard::new(system).0)
```
EdvinNilsson, line 222: I found it a little confusing that SystemExecutionGuard::new(system) did not follow the convention of return a SystemExecutionGuard, but rather a (SystemExecutionGuard, Receiver)-tuple.

crates/scheduler/tests/worker_pool_tests.rs
```diff
@@ -0,0 +1,178 @@
+use crossbeam::channel::unbounded;
+use crossbeam::sync::Parker;
+//noinspection RsUnusedImport - CLion can't recognize that `Write` is being used.
```
EdvinNilsson, line 3: My CLion installation had no problem detecting that Write and timeout were being used when I removed noinspection.

EdvinNilsson (approved):
LGTM

crates/ecs/src/lib.rs
martinjonsson01: Do you think renaming to something other than `new` would help clarify? Maybe `create` or something like that.

crates/scheduler/tests/worker_pool_tests.rs
martinjonsson01: Linux supremacy? 😩 

or, it might be because I've got a bunch of preview features enabled to make it better analyze procedural macros.

crates/ecs/src/lib.rs
EdvinNilsson: Yes, that would be clearer.

JacobBredin (approved):
LGTM!

crates/scheduler/src/executor/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+        self.uid == other.uid
+    }
+}
+
+impl<'a> Task<'a> {
+    #[instrument(skip_all)]
+    fn new(function: impl FnOnce() + Send + 'a) -> Self {
+        static TASK_COUNT: AtomicUsize = AtomicUsize::new(0);
+        // This might overflow and wrap if the application runs for long with many systems,
+        // but that is okay because by then any tasks using the same IDs will be long gone.
```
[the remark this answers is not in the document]
martinjonsson01: Maybe someone wants to run an application at 10,000 FPS with 10,000,000 systems? Who knows? 😝

PR #51, opened 2023-03-20
43 schedule generation

### LINKED ISSUE #43: A schedule that orders systems in a correct manner is generated.

martinjonsson01:
Implements automatic system scheduling based on component accesses. A schedule looks like this:

The way to interpret these diagrams is that an arrow from node `a` to node `b` means that a system called `a` should execute _before_ a system called `b`. You "follow the arrows forward" to execute the schedule.

The scheduler will automatically determine that systems `[read_b, read_a, read_c]`, `[write_b, read_a_write_c]`, `[write_a]` and `[write_ab]` in the above example can run concurrently. Note that this is a _stageless_ scheduler, meaning if `read_b` takes longer to execute than `read_c` then `read_a_write_c` will still get to execute as soon as `read_c` finishes.

>**Note to reviewers:** it might seem like a lot of code but a large amount of it is actually just different test cases. This is because the scheduler is quite difficult to get right (many edge cases), so we need to be thorough with testing. If it feels like a lot to read I'd suggest that you read through one (randomly picked) test of each type and skip the rest as they're quite similar. 
>
> I also suggest that you print out the schedules in each test you look at, it makes understanding them a whole lot easier!
>
> It is very possible I've missed an edge case here and there, so please let me know even if you only have a suspicion that something might be wrong.

To visualize generated schedules, simply print them: 
```rust
println!("schedule is =\n{schedule:#}")
```
and you'll get
```
schedule is =
digraph {
    0 [ label = "read_H_E_write_E_B_G" ]
    1 [ label = "read_A_D_write_B_C" ]
    2 [ label = "read_B_F_D_write_G_E_H" ]
    3 [ label = "read_D_C_B_write_A" ]
    4 [ label = "read_G_write_B_A" ]
    5 [ label = "read_D_C_write_D_C_E" ]
    6 [ label = "read_E_write_E" ]
    7 [ label = "write_A" ]
    1 -> 0 [ label = "0" ]
    2 -> 0 [ label = "0" ]
    2 -> 1 [ label = "0" ]
    3 -> 0 [ label = "0" ]
    3 -> 1 [ label = "0" ]
    4 -> 0 [ label = "0" ]
    4 -> 1 [ label = "0" ]
    4 -> 2 [ label = "0" ]
    4 -> 3 [ label = "0" ]
    5 -> 0 [ label = "0" ]
    5 -> 1 [ label = "0" ]
    5 -> 2 [ label = "0" ]
    3 -> 5 [ label = "0" ]
    6 -> 0 [ label = "0" ]
    6 -> 2 [ label = "0" ]
    6 -> 5 [ label = "0" ]
    7 -> 3 [ label = "0" ]
    7 -> 4 [ label = "0" ]
}
```
which can be input into [viz-js](http://viz-js.com/) to generate:

Closes #43.


JacobBredin (approved):
LGTM! Do not have anything to comment on. 

Sorry for the long review time. The first time I read through the PR I missed the file schedule/mod.rs because GitHub hid it 😤
It made things very confusing.

crates/scheduler/src/schedule/mod.rs
```diff
@@ -0,0 +1,910 @@
...
+/// Whether a precedence graph operation succeeded.
+type PrecedenceGraphResult<T, E = PrecedenceGraphError> = Result<T, E>;
+
+/// Orders [`System`]s based on their precedence.
+#[derive(Default, Clone)]
+pub struct PrecedenceGraph<'systems> {
+    dag: SysDag<'systems>,
+    /// Systems which have been given out, and are awaiting execution.
+    ///
+    /// Warning: Node indices are _not_ stable and will be invalidated if [`dag`] is mutated.
```
EdvinNilsson, line 60: The intra-doc link ``[`dag`]`` is broken because `dag` is not in scope. This can be fixed by replacing it to ``[`dag`](Self::dag)``.

crates/scheduler/src/schedule/mod.rs
```diff
@@ -0,0 +1,910 @@
...
+}
+
+impl<'systems> Schedule<'systems> for PrecedenceGraph<'systems> {
+    fn generate(systems: &'systems [Box<dyn System>]) -> ScheduleResult<Self> {
+        let mut dag = Dag::new();
+
+        for system in systems.iter().map(|system| system.as_ref()) {
+            let node = dag.add_node(system);
+
+            // Find nodes which current system must run _after_ (i.e. current has dependency on).
```
EdvinNilsson, line 107: Why use markdown in normal comments when it is only supported in doc comments?

crates/scheduler/src/schedule/mod.rs
```diff
@@ -0,0 +1,910 @@
...
+    ///
+    /// Warning: Node indices are _not_ stable and will be invalidated if [`dag`] is mutated.
+    already_executed: Vec<NodeIndex>,
+}
+
+impl<'systems> Debug for PrecedenceGraph<'systems> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        // Instead of printing the struct, format the DAG in dot-format and print that.
+        // (so it can be viewed through tools like http://viz-js.com/)
+        writeln!(f, "{:?}", daggy::petgraph::dot::Dot::new(self.dag.graph()))
```
EdvinNilsson, line 72: A get alot of false positives in this file from daggy in CLion. Are you also experiencing this? Here I get:
```
`Dot<&DiGraph<Sys, i32>>` doesn't implement `Debug` (required by {:?})
```

EdvinNilsson (approved):
LGTM

crates/scheduler/src/schedule/mod.rs
martinjonsson01: Neither seem to work for me, so I assumed it was impossible to have it correctly link. I'm only trying it inside CLion though, do you mean that `[`dag`](Self::dag)` would correctly link it if we were to generate out the documentation?

crates/scheduler/src/schedule/mod.rs
martinjonsson01: To place extra emphasis on specific words, to make it clear for the reader that I want to stress a point in this comment.

crates/scheduler/src/schedule/mod.rs
martinjonsson01: Yup, it's horrible... 

I searched for solutions and found a bunch of bug reports about it, but they were all marked resolved so we must be encountering a new case. 

I tried looking into ways of disabling that error for this file, but it seems that there's no way to disable errors using comments and the like, only warnings. :(

crates/scheduler/src/schedule/mod.rs
EdvinNilsson: When using ``[`dag`](Self::dag)`` the warning is gone when running `cargo doc`, but it looks like CLion still can't follow the link.

martinjonsson01: Ah, I usually don't run `cargo doc` but that's good to know! I'll update it then for the sake of being correct. 👍

PR #64, opened 2023-03-21
40 integrate with tracy profiling tool

### LINKED ISSUE #40: Integrate with Tracy profiling tool
## Reason
To be able to diagnose and showcase the performance of the MVP, allow the Tracy profiler to connect to a running instance of the engine.

## Requirements
- [x] The [Tracy profiling tool](https://github.com/wolfpld/tracy) can connect to the engine.
- [x] Diagnostic data such as function calls is shown in the tool.
- [x] Frames are distinctly visible in the profiler.
- [x] Sub-frames (i.e. batch of systems executing concurrently) are visible in the profiler.
- [x] Profiling is not necessary for the engine to run.
- [x] Profiling does not impact performance when turned off.
  - [x] Profiling has to be enabled with the feature-flag `profile`.

## Dependencies
- [x] #36 
- [x] #39

martinjonsson01:
Enables profiling integration with the [Tracy profiling tool](https://github.com/wolfpld/tracy).

image

[Demo video](https://i.imgur.com/ELvjZkN.mp4)

Profiling is not enabled by default, but is an optional feature that can be enabled using a compilation flag and a configuration call to `Application`.

When profiling is enabled, no logging to console is done. This is to increase the performance. 

To run the profiler on the `parallel.rs` example in the `scheduler` crate:
1. Switch out the call to `.with_tracing()?` to `with_profiling()?`: 
```rust 
let mut app = Application::default()
        //.with_tracing()?
        .with_profiling()?
```
3. Enable the `profile` feature by adding the compilation flag `--features profile` to the end of the `cargo run` command (either in CLion run configuration, or in terminal).
4. Start Tracy.
5. Run `parallel.rs`.
6. In Tracy, click "Connect".

Closes #40

### DIFF
--- a/crates/ecs/src/lib.rs
+++ b/crates/ecs/src/lib.rs
@@ -28,6 +28,7 @@
 )]
 
 pub mod logging;
+mod profiling;
 
 use crate::ApplicationError::ScheduleGeneration;
 use crossbeam::channel::{bounded, Receiver, Sender, TryRecvError};
--- a/crates/ecs/src/lib.rs
+++ b/crates/ecs/src/lib.rs
@@ -489,10 +489,10 @@ impl<Function: Send + Sync, Parameters: SystemParameters> Debug
             parameter_names_text.push_str(parameter_name);
         }
 
-        writeln!(f, "FunctionSystem {{")?;
-        writeln!(f, "    system = {}", &self.function_name)?;
-        writeln!(f, "    parameters = {parameter_names_text}")?;
-        writeln!(f, "}}")
+        f.debug_struct("FunctionSystem")
+            .field("system", &self.function_name)
+            .field("parameters", &parameter_names_text)
+            .finish()
     }
 }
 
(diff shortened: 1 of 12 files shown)


EdvinNilsson (approved):
Looks good! Probably just a limitation of Tracy, but the profiler becomes unusably slow when the application runs at 1000+ fps in release builds. You can still profile release builds by just adding a system that blocks the update loop to lower the fps.

crates/ecs/src/lib.rs
```diff
@@ -488,10 +489,10 @@ impl<Function: Send + Sync, Parameters: SystemParameters> Debug
...
         }
 
-        writeln!(f, "FunctionSystem {{")?;
-        writeln!(f, "    system = {}", &self.function_name)?;
-        writeln!(f, "    parameters = {parameter_names_text}")?;
-        writeln!(f, "}}")
+        f.debug_struct("FunctionSystem")
+            .field("system", &self.function_name)
+            .field("parameters", &parameter_names_text)
+            .finish()
```
JacobBredin, line 495: Fancy ✅

JacobBredin (approved):
LGTM! 

It runs on my machine so I'm happy 👍

PR #65, opened 2023-03-22
54 system tick barrier

### LINKED ISSUE #54: System tick barrier
## Reason
To avoid systems without dependencies from running more times per tick than all the others, there needs to be a synchronization barrier at the end of every tick such that no more systems are executed until all have executed once.

## Requirements
- [ ] A system with no dependencies only runs once per tick
- [ ] The scheduler is still stageless, i.e. systems that execute faster can still move forward _until they reach the tick barrier_

## Dependencies
- [ ] #36

martinjonsson01:
It appears I had already implemented this without knowing, so I've just added a test to ensure that it stays that way.

Closes #54

### DIFF
--- a/crates/scheduler/src/schedule/mod.rs
+++ b/crates/scheduler/src/schedule/mod.rs
@@ -341,6 +341,8 @@ mod tests {
     use itertools::Itertools;
     use ntest::timeout;
     use proptest::prop_assume;
+    use std::thread;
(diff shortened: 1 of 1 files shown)


JacobBredin (approved):
LGTM! Approved ✅

EdvinNilsson (approved):
LGTM

Issue #72, opened 2023-03-24
Dynamic Archetypes


JacobBredin:
## Reason
Adding and removing components to entities should move enities and their components between archetypes that store these types.

## Requirements
- [ ] First requirement
- [ ] Second requirement
- [ ] Third requirement

## Dependencies
- [ ] \#issue number

Issue #53, opened 2023-03-20
System command buffers


martinjonsson01:
## Reason
To allow systems to add and remove components and entities during runtime, they should have access to a command buffer which they can input commands into.

## Requirements
- [x] The following commands are supported:
  - [x] Add component X to entity Y
  - [x] Remove component X from entity Y
  - [x] Remove entity Y
  - [x] Create entity with components A, B, C, ...
- [x] Command buffers are able to be written to concurrently (i.e. each system has their own buffer)
- [x] Command buffers are optional, systems don't have to have one
- [ ] There is a command buffer _system_, which:
  - [x] is automatically inserted into the end of every frame;
  - [x] executes the commands;
  - [x] could theoretically (or actually) be inserted in several places in the schedule, to give multiple sync-points.

## Dependencies
- [x] #36
- [x] #38
- [x] #72 
- [x] #66

PR #77, opened 2023-03-26
38 Implement entity archetypes in MVP

JacobBredin:
An archetype data structure has been implemented with basic functionality for adding entities, adding and removing components from entities and adding component vecs to archetypes exists.

Application and world has been rewritten to use archetypes, but only in a limited way. Currently there only exists one archetype which I have aptly named "Big Archetype", that contain all entities and their components. This archetype is located on index 0 of the `Archetypes: Vec<Archetype>`. Methods for fetching multiple read or write component vecs from archetypes have been implemented and will be compatible with future inclusion of storing multiple `Archetypes` in `World`. 

I've also conducted some refactoring, moving methods for handling "raw" component vecs to their own method, these are: `create_raw_component_vec`, `borrow_component_vec` and `borrow_component_vec_mut`.

Special thanks to @EdvinNilsson for helping me getting over some problems concerning integrating `Archetypes` with the macros and `SystemParameter` traits.


crates/ecs/src/lib.rs
```diff
@@ -585,27 +871,51 @@ impl<'a, Component> Deref for Read<'a, Component> {
...
-        if let Some(component_vec) = borrowed {
-            if let Some(Some(component)) = component_vec.get(entity.id) {
-                return Some(Self {
-                    // The caller is responsible to only use the
-                    // returned value when BorrowedData is still in scope.
-                    #[allow(trivial_casts)]
-                    output: &*(component as *const Component),
-                });
+    unsafe fn fetch_parameter(borrowed: &mut Self::BorrowedData<'_>) -> Option<Option<Self>> {
+        if let Some(archetype) = (borrowed.1).get_mut(borrowed.0) {
```
martinjonsson01, line 899: when working with tuples it's usually best to destructure them ASAP, to avoid obtuse `.0`s and `.1`s everywhere.

I.e.
```rust
let (archetype_index, component_vec) = borrowed;
```
and then you use `archetype_index` and `component_vec` instead.

crates/ecs/src/lib.rs
```diff
@@ -585,27 +871,51 @@ impl<'a, Component> Deref for Read<'a, Component> {
...
+    tmp
+}
 
-    fn borrow(world: &World) -> Self::BorrowedData<'_> {
-        world.borrow_component_vec::<Component>()
+impl<Component: Debug + Send + Sync + 'static + Sized> SystemParameter for Read<'_, Component> {
+    type BorrowedData<'components> = (
+        usize,
+        Vec<(usize, ReadComponentVec<'components, Component>)>,
+    );
```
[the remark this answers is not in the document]
martinjonsson01: An alternative to this is destructuring, like I wrote in the comment below. That would be enough if you also use newtypes in order to disambiguate what `usize` in the tuple means.

crates/ecs/src/lib.rs
```diff
@@ -585,27 +871,51 @@ impl<'a, Component> Deref for Read<'a, Component> {
...
+                return if let Some(component) = component_vec.get(get_and_inc(&mut archetype.0)) {
+                    if let Some(component) = component {
+                        return Some(Some(Self {
+                            // The caller is responsible to only use the
+                            // returned value when BorrowedData is still in scope.
+                            #[allow(trivial_casts)]
+                            output: &*(component as *const Component),
+                        }));
+                    }
+                    Some(None)
```
martinjonsson01, line 910: this is a code smell. If you're working with multiple `Option`s then you should chain them together using combinators, not wrap them inside of each other!

crates/ecs/src/lib.rs
```diff
@@ -585,27 +871,51 @@ impl<'a, Component> Deref for Read<'a, Component> {
...
-                return Some(Self {
-                    // The caller is responsible to only use the
-                    // returned value when BorrowedData is still in scope.
-                    #[allow(trivial_casts)]
-                    output: &*(component as *const Component),
-                });
+    unsafe fn fetch_parameter(borrowed: &mut Self::BorrowedData<'_>) -> Option<Option<Self>> {
+        if let Some(archetype) = (borrowed.1).get_mut(borrowed.0) {
+            if let Some(component_vec) = &archetype.1 {
+                return if let Some(component) = component_vec.get(get_and_inc(&mut archetype.0)) {
```
martinjonsson01, line 901: I thought you would use `get_and_inc` in some elegant iterator-chain, and that's why you needed it, but this usage here is quite confusing imo. 

Why not simply use `archetype.0` here and then inside of the if-clause do `archetype.0 += 1` like you do in the else-clause?

crates/ecs/src/lib.rs
```diff
@@ -614,27 +924,45 @@ impl<Component: Send + Sync + 'static + Sized> SystemParameter for Read<'_, Comp
...
+                            #[allow(trivial_casts)]
+                            output: &mut *(component as *mut Component),
+                        }));
+                    }
+                    Some(None)
+                } else {
+                    // End of archetype
+                    borrowed.0 += 1;
+                    Self::fetch_parameter(borrowed)
+                };
```
martinjonsson01, line 962: see comments in Read-impl

crates/ecs/src/lib.rs
```diff
@@ -806,4 +1138,391 @@ mod tests {
...
+        let result_u32 = archetype.borrow_component_vec::<u32>().unwrap();
+        let result_u64 = archetype.borrow_component_vec::<u64>().unwrap();
+
+        // entity 1 should have index 0, as it was added first
+        assert_eq!(result_u32.get(0).unwrap().unwrap(), 21);
+        assert_eq!(result_u64.get(0).unwrap().unwrap(), 212);
+
+        // entity 2 should have index 1 as it was added second
+        assert_eq!(result_u32.get(1).unwrap().unwrap(), 35);
+        assert_eq!(result_u64.get(1).unwrap().unwrap(), 123);
```
martinjonsson01, line 1176: nit:
can be simplified
```suggestion
        assert_eq!(vec![Some(21), Some(35)], *result_u32);
        assert_eq!(vec![Some(212), Some(123)], *result_u64);
```

crates/ecs/src/lib.rs
```diff
@@ -806,4 +1138,391 @@ mod tests {
...
+    fn components_are_stored_on_correct_index_independent_of_insert_order() {
+        let mut archetype = Archetype::default();
+
+        let entity_1_id = 0;
+        let entity_2_id = 10;
+        let entity_3_id = 5;
+
+        let entity_1_idx = 0;
+        let entity_2_idx = 1;
+        let entity_3_idx = 2;
```
martinjonsson01, line 1189: nit:
I assume these are indices and are not meant to be read as "identifier x"? If so, you could probably afford the extra characters to just write out `entity_n_index`

crates/ecs/src/lib.rs
```diff
@@ -806,4 +1138,391 @@ mod tests {
...
+        world.store_entity_in_archetype(e3.id, 1);
+
+        // insert some components...
+        world.add_component::<u64>(e1.id, 1);
+        world.add_component::<u32>(e1.id, 2);
+
+        world.add_component::<u64>(e2.id, 3);
+        world.add_component::<u32>(e2.id, 4);
+
+        world.add_component::<u32>(e3.id, 6);
```
martinjonsson01, line 1403: since the arrangement of these two tests are the exact same, you should extract that into a common setup-function. This will help us a lot down the line when we refactor the implementation and therefore also will need to refactor the tests.

crates/ecs/src/lib.rs
```diff
@@ -806,4 +1138,391 @@ mod tests {
...
+
+        let mut result: Vec<u32> = vec![];
+
+        let mut borrowed = <Read<u32> as SystemParameter>::borrow(
+            &world,
+            &[<Read<u32> as SystemParameter>::signature()],
+        );
+        unsafe {
```
martinjonsson01, line 1411: you should add a [`// SAFETY: `](https://std-dev-guide.rust-lang.org/documentation/safety-comments.html#inside-safe-elements) comment explaining why it's safe to use the functions this way.

crates/ecs/src/lib.rs
```diff
@@ -806,4 +1138,391 @@ mod tests {
...
+        archetype.add_component_vec::<u32>();
+        archetype.add_component_vec::<f32>();
+
+        archetype.add_component::<u32>(10, 1);
+        archetype.add_component::<u32>(20, 2);
+        archetype.add_component::<u32>(30, 3);
+
+        archetype.add_component::<f32>(10, 1.0);
+        archetype.add_component::<f32>(20, 2.0);
+        archetype.add_component::<f32>(30, 3.0);
```
martinjonsson01, line 1484: nit:
I think I preferred the way you handled it in other tests, where you named the entity indices `entity0_index`, `entity1_index` and so on... instead of using the indices directly like this (i.e. magic numbers). Using named variables makes it a lot clearer that this `10` is the same thing as this other `10`.

crates/ecs/src/lib.rs
```diff
@@ -806,4 +1138,391 @@ mod tests {
...
+    #[test_case(vec![Vec::<usize>::new(), Vec::<usize>::new(), Vec::<usize>::new(), Vec::<usize>::new()], vec![]; "multiple empty")]
+    #[test_case(vec![Vec::<usize>::new(), vec![1,2,3,4]], vec![]; "one empty, one not")]
+    #[test_case(vec![vec![2,1,1,1,1], vec![1,1,1,1,2], vec![1,1,2,1,1]], vec![2,1,1,1,1]; "multiple of the same number")]
+    fn intersection_returns_expected_values(
+        test_vecs: Vec<Vec<usize>>,
+        expected_value: Vec<usize>,
+    ) {
+        // Construct test values, to avoid upsetting Rust and test_case
+        let borrowed_test_vecs: Vec<&Vec<usize>> = test_vecs.iter().collect();
+        let borrowed_expected_value: Vec<&usize> = expected_value.iter().collect();
```
martinjonsson01, line 1520: nit:
```suggestion
        // Borrow inner values, because `intersection` expects a vector of references.
        let borrowed_test_vecs: Vec<_> = test_vecs.iter().collect();
        let borrowed_expected_value: Vec<_> = expected_value.iter().collect();
```

crates/ecs/src/lib.rs
```diff
@@ -806,4 +1138,391 @@ mod tests {
...
+
+    // Intersection tests:
+    #[test_case(vec![vec![1,2,3]], vec![1,2,3]; "self intersection")]
+    #[test_case(vec![vec![1,2,3], vec![1,2,3]], vec![1,2,3]; "two of the same")]
+    #[test_case(vec![vec![1], vec![2,3], vec![4]], vec![]; "no overlap, no matches")]
+    #[test_case(vec![vec![1,2], vec![2,3], vec![3,4]], vec![]; "some overlap, no matches")]
+    #[test_case(vec![vec![1,2,3,4], vec![2,3], vec![3,4]], vec![3]; "some matches")]
+    #[test_case(vec![Vec::<usize>::new()], vec![]; "empty")]
+    #[test_case(vec![Vec::<usize>::new(), Vec::<usize>::new(), Vec::<usize>::new(), Vec::<usize>::new()], vec![]; "multiple empty")]
+    #[test_case(vec![Vec::<usize>::new(), vec![1,2,3,4]], vec![]; "one empty, one not")]
```
martinjonsson01, line 1512: nit:
```suggestion
    #[test_case(vec![vec![]], vec![]; "empty")]
    #[test_case(vec![vec![], vec![], vec![], vec![]], vec![]; "multiple empty")]
    #[test_case(vec![vec![], vec![1,2,3,4]], vec![]; "one empty, one not")]
```

martinjonsson01 (approved):
Nice work! I've got some questions and suggestions concerning code quality improvements, though.

Overall I think it's a well thought-out implementation, and it could therefore technically be merged (which is why I'm approving). But I would like to see many of my concerns addressed to make the code easier to work with during the rest of engine development. 

If you want to wait with some of the improvement suggestions until you work on #72, then that's fine with me as long as you leave a to-do comment in those places.

crates/ecs/src/lib.rs
```diff
@@ -265,74 +259,293 @@ impl<'systems> Schedule<'systems> for Unordered<'systems> {
...
+}
+
+impl Archetype {
+    /// Adds an `entity_id` to keep track of and store components for.
+    fn store_entity(&mut self, entity_id: usize) {
+        let entity_index = self.entity_id_to_component_index.len();
+
+        self.component_typeid_to_component_vec
+            .values_mut()
+            .for_each(|v| v.push_none());
```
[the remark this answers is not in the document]
JacobBredin: It is a temporary solution. But I'm not sure how it can be solved. Imagine that you store an entity in an archetype, but you have not yet transferred all components to the archetypes respective component vectors. All indices need to line up for each entity, but if you only add some of the components the indices will quickly become uneven. If we are to remove Option, we need some way of guaranteeing that the indices do not become unaligned.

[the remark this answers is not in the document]
martinjonsson01: >Imagine that you store an entity in an archetype, but you have not yet transferred all components to the archetypes respective component vectors.

What do you mean "have not yet transferred all components"? Do you mean that if I add an entity with components `A`, `B` and `C` there will be some nanoseconds between when I add `A`, `B` and `C` to their respective component vecs? That shouldn't be an issue since mutating an archetype like that should require exclusive access to it, right?

crates/ecs/src/lib.rs
```diff
@@ -265,74 +259,293 @@ impl<'systems> Schedule<'systems> for Unordered<'systems> {
...
+        if let Some(&index) = self.entity_id_to_component_index.get(&entity_id) {
+            self.component_typeid_to_component_vec
+                .values()
+                .for_each(|vec| vec.swap_remove(index));
+            self.entity_id_to_component_index.remove(&entity_id);
+            // update index of compnonets of entity on last index
+            self.entity_id_to_component_index
+                .insert(self.last_entity_id_added, index);
+        }
+    }
```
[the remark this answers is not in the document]
JacobBredin: Seems a bit overkill, but I can add it

crates/ecs/src/lib.rs
```diff
@@ -265,74 +259,293 @@ impl<'systems> Schedule<'systems> for Unordered<'systems> {
...
+    fn add_component<ComponentType: Debug + Send + Sync + 'static>(
+        &mut self,
+        entity_id: usize,
+        component: ComponentType,
+    ) {
+        if let Some(&entity_index) = self.entity_id_to_component_index.get(&entity_id) {
+            if let Some(mut component_vec) = self.borrow_component_vec_mut::<ComponentType>() {
+                component_vec[entity_index] = Some(component);
+            }
+        }
```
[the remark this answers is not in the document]
JacobBredin: Yeah, makes it easier to find the bugs 🪲

crates/ecs/src/lib.rs
```diff
@@ -265,74 +259,293 @@ impl<'systems> Schedule<'systems> for Unordered<'systems> {
...
+    }
+
+    /// Adds a component vec of type `ComponentType` if no such vec already exists.
+    fn add_component_vec<ComponentType: Debug + Send + Sync + 'static>(&mut self) {
```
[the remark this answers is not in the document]
JacobBredin: I thought it might be a useful feature for different applications of using Archtypes. Maybe it should be moved to a builder struct. Currently I use it to implement the BigArchtype. I'll add a comment about #72 needing to evaluate if it should stay.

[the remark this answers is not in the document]
martinjonsson01: 👍 

I like the idea of moving it to a builder struct, since then it's clear that you only add new component vecs during a construction-phase, and then once you've built the archetype it is "immutable in it's signature".

crates/ecs/src/lib.rs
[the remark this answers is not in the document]
JacobBredin: Nevermind, its to do with `BigArchetype`, but it needs to exist right now, but should be removed in #72. I'll add a commet

crates/ecs/src/lib.rs
```diff
@@ -265,74 +259,293 @@ impl<'systems> Schedule<'systems> for Unordered<'systems> {
...
+        {
+            return borrow_component_vec_mut(component_vec.as_ref());
+        }
+        None
+    }
+
+    /// Adds a component of type `ComponentType` to the specified `entity`.
+    fn add_component<ComponentType: Debug + Send + Sync + 'static>(
+        &mut self,
+        entity_id: usize,
```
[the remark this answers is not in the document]
JacobBredin: Storing entities in a HashMap will make it impossible to find them, since you would need to know which generation each entity has. So the idea was to only use entity in `World`. However, this might need to change when entities can be added and removed in a more "official" way.

I'll add a todo comment about this for #72.

crates/ecs/src/lib.rs
```diff
@@ -1455,7 +1453,7 @@ mod tests {
         let vecs_u32 = world
             .borrow_component_vecs_with_signature::<u32>(&[TypeId::of::<u32>()])
             .unwrap();
-
+        eprintln!("vecs_u32 = {:#?}", vecs_u32);
```
martinjonsson01, line 1456: just a small tip, you can inline printable variables as long as they're not a more complex expression:
```suggestion
        eprintln!("vecs_u32 = {vecs_u32:#?}");
```

crates/scheduler/examples/parallel.rs
```diff
@@ -16,7 +16,7 @@ fn main() -> Result<(), Report> {
         .add_system(read_write_many);
 
     for i in 0..20 {
-        let entity = app.create_entity();
+        let entity = app.create_entity().unwrap();
```
martinjonsson01, line 19: while it's fine to use `.unwrap()` in tests (since they should fail with a panic if something's wrong), our example code should be written much more nicely and be exemplary in its handling of errors (i.e. propagating them instead of panicking immediately)
```suggestion
        let entity = app.create_entity()?;
```

crates/ecs/src/lib.rs
```diff
@@ -265,74 +259,293 @@ impl<'systems> Schedule<'systems> for Unordered<'systems> {
...
-                // Panicking helps us detect errors in the scheduling algorithm more quickly.
-                return match component_vec.try_write() {
-                    Ok(component_vec) => Some(component_vec),
-                    Err(TryLockError::WouldBlock) => panic_locked_component_vec::<ComponentType>(),
-                    Err(TryLockError::Poisoned(_)) => panic!("Lock should not be poisoned!"),
-                };
-            }
+    fn add_component<ComponentType: Debug + Send + Sync + 'static>(
+        &mut self,
+        entity_id: usize,
```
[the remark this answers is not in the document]
JacobBredin: Because archetypes should not know about generation, they are only interested in entity Id. Adding Generation was necessary as it will not be needed until #53.  I've added a comment about this in the entity struct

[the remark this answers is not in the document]
martinjonsson01: I see, but is it worth it to not use our specific `Entity` type whose purpose is to represent a unique entity? For sure, the generation might be unnecessary in some functions, but does it matter that it's present? 

I'd argue that it's both more consistent and more readable to always use `Entity`s and not break them up into their constituent parts (again, this goes back to the problem of using `usize` everywhere for different things). 

Or is there another reason not to simply pass in `Entity`, other than on principle of it not being necessary? A suggestion is to make the `id` field public and the `generation` field private, so code outside of the module isn't even aware of the `generation`.

crates/ecs/src/lib.rs
```diff
@@ -585,27 +978,46 @@ impl<'a, Component> Deref for Read<'a, Component> {
...
+        Vec<(ComponentIdx, ReadComponentVec<'components, Component>)>,
+    );
+
+    fn borrow<'a>(
+        world: &'a World,
+        signature: &[TypeId],
+    ) -> SystemParameterResult<Self::BorrowedData<'a>> {
+        let component_vecs = world
+            .borrow_component_vecs_with_signature::<Component>(signature)
+            .map_err(SystemParameterError::BorrowComponentVecs)?;
```
EdvinNilsson, line 993: There should not be a runtime error when a system is querying a component type that has not been added to any entity. The component may be added later (when it is possible to add/remove components in runtime).

crates/scheduler/examples/parallel.rs
martinjonsson01: > while it's fine to use .unwrap() in tests (since they should fail with a panic if something's wrong)

Or what do you mean?

crates/ecs/src/lib.rs
```diff
@@ -68,16 +72,12 @@ pub struct Application {
...
-            id: self.world.entities.len(),
-            _generation: 0,
-        };
-        self.world.entities.push(entity);
-        entity
+    pub fn create_entity(&mut self) -> AppResult<Entity> {
+        let entity = self.world.create_new_entity();
+        self.world
+            .store_entity_in_archetype(entity.id, 0)
+            .map_err(ApplicationError::World)?; // todo(#72): Change so that all entities are not stored in the same big archetype, that is change index 0 to be something else.
```
EdvinNilsson, line 79: Should we propagate implementation errors to the end user? There is no way for the end user to recover from them.

crates/scheduler/examples/parallel.rs
JacobBredin: I thought this was in the ecs/lib test, my bad

EdvinNilsson (approved):
LGTM

crates/ecs/src/lib.rs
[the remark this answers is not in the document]
JacobBredin: Perhaps using EntityId(usize) as its own type would make it easier to understand the HashMaps that use <usize, usize>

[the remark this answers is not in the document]
martinjonsson01: Yeah, that'd be a lot better. 

But then what's the purpose of having two different entity types? `EntityId` and `Entity` represent the exact same thing, but one contains a smudge more information.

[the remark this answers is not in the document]
JacobBredin: Because you cannot use `Entity` in a HashMap, they will be impossible to find, thats why you want to map EntityId->ComponentIndex for example. But I think we can make these changes later, refactoring can happen in stages right?

crates/ecs/src/lib.rs
JacobBredin: World has two types of errors: input errors and internal data structure errors. The first one can be caused by trying to borrow a component vec of a type that is not stored, these should be propagated. The other kind should just panic or we should add code for cleaning up the data structures.

crates/ecs/src/lib.rs
JacobBredin: I was also thinking about this. In this case this error is only thrown if no archetypes exists of associated with a type. But this should only result in no archetypes being returned

crates/ecs/src/lib.rs
```diff
@@ -1332,82 +1439,92 @@ mod tests {
...
             .collect();
-        println!("{:?}", result);
+        println!("{result:?}");
 
         assert_eq!(result, vec![2, 4, 6])
     }
 
     #[test]
-    fn querying_with_archetypes() {
+    fn borrowing_none_existant_component_returns_empty_vec() {
```
martinjonsson01, line 1488: nit:
```suggestion
    fn borrowing_non_existent_component_returns_empty_vec() {
```

crates/ecs/src/lib.rs
```diff
@@ -265,74 +259,293 @@ impl<'systems> Schedule<'systems> for Unordered<'systems> {
...
+        if let Some(&entity_index) = self.entity_id_to_component_index.get(&entity_id) {
+            if let Some(mut component_vec) = self.borrow_component_vec_mut::<ComponentType>() {
+                component_vec[entity_index] = Some(component);
+            }
+        }
+    }
+
+    /// Adds a component vec of type `ComponentType` if no such vec already exists.
+    fn add_component_vec<ComponentType: Debug + Send + Sync + 'static>(&mut self) {
+        if !self.contains::<ComponentType>() {
```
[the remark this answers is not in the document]
JacobBredin: They both return result types now. add_component_vec is idempotent, adding multiple of the same type changes nothing.

crates/ecs/src/lib.rs
```diff
@@ -265,74 +259,293 @@ impl<'systems> Schedule<'systems> for Unordered<'systems> {
...
+
+    fn borrow_component_vecs_with_signature_mut<ComponentType: Debug + Send + Sync + 'static>(
+        &self,
+        signature: &[TypeId],
+    ) -> Vec<WriteComponentVec<ComponentType>> {
+        let archetype_indices = self.get_archetype_indices(signature);
+        self.borrow_component_vecs_mut(&archetype_indices)
+    }
+
+    fn get_archetype_indices(&self, signature: &[TypeId]) -> Vec<&usize> {
```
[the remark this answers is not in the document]
JacobBredin: I now use these three instead of usize 👍

crates/ecs/src/lib.rs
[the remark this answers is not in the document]
JacobBredin: fixed 🔧🪛

crates/ecs/src/lib.rs
```diff
@@ -1265,9 +1294,18 @@ mod tests {
     fn components_are_stored_on_correct_index_independent_of_insert_order() {
         let mut archetype = Archetype::default();
 
-        let entity_1_id = 0;
-        let entity_2_id = 10;
-        let entity_3_id = 5;
+        let entity_1 = Entity {
+            id: 0,
+            _generation: 0,
+        };
```
martinjonsson01, line 1300: Looks like a constructor `Entity::with_id(n)` would be handy? If this'll be unnecessary in the very next PR then you can skip this, otherwise it could be nice to have.

martinjonsson01 (approved):
LGTM!

crates/ecs/src/lib.rs
JacobBredin: I'll place it in the test module, since it will not be used anywhere else

PR #78, opened 2023-03-29
48 System schedule reordering

### LINKED ISSUE #48: System schedule reordering
## Reason
In order to reduce the total makespan of the schedule, thereby increasing the number of systems that are executed in parallel, the schedule should reorder some systems in some cases.

Example of what we would like:

Example of what it currently schedules:

## Requirements
- [ ] The currently ignored test `schedule_reorders_systems_to_reduce_makespan` passes.

## Dependencies
- [x] #43

ChristofferPerssons:
Seemingly functioning solution for makespan minimization . No proof actually found however the algorithm has been checked on some test graphs manually and seems to function well there. 

If at any point the application panics from the expect() in the genereration phase, the algorithm can probably be considered broken and another solution is needed.

Closes #48

### DIFF
--- a/crates/scheduler/src/schedule/mod.rs
+++ b/crates/scheduler/src/schedule/mod.rs
@@ -13,7 +13,10 @@ use crate::schedule::PrecedenceGraphError::{
 };
 use crossbeam::channel::{Receiver, RecvError, Select};
 use daggy::petgraph::dot::{Config, Dot};
-use daggy::petgraph::visit::{IntoNeighbors, IntoNeighborsDirected, IntoNodeIdentifiers};
+use daggy::petgraph::prelude::EdgeRef;
+use daggy::petgraph::visit::{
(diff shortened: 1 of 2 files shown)


crates/scheduler/src/schedule/mod.rs
```diff
@@ -628,7 +669,7 @@ mod tests {
     }
 
     #[test]
-    #[ignore] // todo(#48): This problem has not been solved yet, so test is ignored for now.
+    // todo(#48): This problem has not been solved yet, so test is ignored for now.
```
martinjonsson01, line 672: This comment should be removed, since you've resolved the to-do.

martinjonsson01 (approved):
LGTM!

Matte02 (approved):
As far as I can tell it looks good! Good job!
Seems like a good solution for the problem, and can't find anything obvious for why it would not work.

Issue #19, opened 2023-02-13
Run single system in different threads - Crossbeam


martinjonsson01:
## Reason
As a very basic form of concurrency, run each group of components in its own thread to set up the infrastructure for it.

## Requirements
- [ ] Each group of components is run on a separate thread.
- [ ] All group of components run once, then do not run again until all group of components have finished executing (e.g. a single frame/tick) -- a barrier exists.
- [ ] Threads are reused every tick, new threads are not spawned for every execution of a group of components.

## Dependencies
- [ ] #6

PR #81, opened 2023-03-29
79 extract systems module from lib

martinjonsson01:
Refactors out the code of systems into its own module, to reduce clutter.

The implementations aren't changed, so you just need to pay attention to the general structure, modules and traits.

Note that the module `ecs::systems::iteration` is created because my next step is to implement parallel system iteration.

This is a part of issue #19.


JacobBredin (approved):
LGTM, well needed 👍

Matte02 (approved):
LGTM 👍

Issue #50, opened 2023-03-20
Allow systems to access components from multiple entities at once


EdvinNilsson:
## Reason
This is required for the N-body case study and for the “Foreign entity access” sub-problem.

## Requirements
- [x] Allow systems to access entities other than the one currently being iterated
- [x] There is an ergonomic way of declaring this query
- [x] The scheduler is informed about component accesses

## Dependencies
- [x] #34
- [x] #38

PR #85, opened 2023-04-01
80 parallel system iteration

### LINKED ISSUE #19: Run single system in different threads - Crossbeam

### LINKED ISSUE #80: Parallel system iteration
## Reason
Run batches of entities within a system's execution in parallel, to increase performance when there are many entities.

## Requirements
- [x] Each batch of entities is run on a separate thread using the task system in `WorkerPool`.
- [x] All batches of entities run once, then do not run again until all batches of entities have finished executing (e.g. a single frame/tick) -- a barrier exists.
- [x] This should not happen on low numbers of entitites (i.e. < 1000).
- [x] The cutoff point for how many/how large batches there should be is decided (either statically or dynamically) such that there is a _performance gain_ in running them in parallel. I.e. the overhead of creating the batches and tasks is not greater than the performance gain.

## Dependencies
- [x] #79

martinjonsson01:
Implements intra-system parallelization, meaning now a single system can automatically be split up to execute different parts of its loop concurrently. 

Currently, the hardcoded segment size is `100`, which some preliminary testing has shown works okay for both small and large numbers of entities.

> Note: This branch was created off of `50-foreign-entity-access` and so will have to be merged after #50 is finished.

> The following examples are from the `parallel.rs` example. It has three initial systems which run concurrently, one of which has no parameters meaning it's almost never visible in the graphs.  
image

For example, when the segment size is very small this will result in a huge number of tasks which means the total overhead incurred by task creation will outweigh the performance benefit of running them in parallel:
### 200,000 entities - segment size 20 - 183 ms
entity_200000_segment_20

### 2,000 entities - segment size 10 - 16 ms
entity_2000_segment_10

On the other hand, when the segment size is very large, this will result in a low number of tasks in relation to the number of systems (if the segment size is larger than the number of iterations a system performs, then only a single task is generated like before):

### 2,000 entities - segment size 1000 - 4 ms
Since this example has a segment size half of the number of iterations per system, it shows how much idle cpu time is left on the table when the segment size is too large. Look at the other images and notice how they keep most threads much more occupied.

entity_2000_segment_1000

Both seem to work well with segment size of `100`:

### 200,000 entities - segment size 100 - 169 ms
entity_200000_segment_200

### 2,000 entities - segment size 100 - 2 ms
entity_2000_segment_100

Closes #80 
Closes #19


EdvinNilsson (approved):
Looks good! However, this implementation only executes systems in parallel, while the parameter fetching is still sequential. This works well for more computationally heavy systems, but less well for more lightweight systems where fetching data takes most of the time. 

I compared this implementation with the sequential version with two examples.  

When using a simple system that just moves one million entities, I saw around 2.7 times worse performance with this implementation compared to the sequential version. When I tried this on a more computationally heavy system similar to n-body with 10 000 entities, then I saw around a 5.5 times speedup (with 12 logical cores). 

I know that implementing parallel fetching is a little more tricky, because the different segments will be on multiple archetypes and multiple threads need to access different parts of the same component vector at the same time. I think that implementing this would significantly improve the performance of the weather simulation case-study. Do you think this is worth implementing?

EdvinNilsson:
> I didn't see an easy way to fetch component data in parallel, could you specify what you mean by that? Fetching each parameter in parallel could possibly be done, but it'd still require all parameters to be fetched before iteration could begin. Do you mean that the fetching could be segmented across the entire query, just like how I'm currently segmenting the iteration? I.e. fetch parameters for entities 0..100 and running them, in parallel with fetching parameters for entities 100..200 and running them?

To fetch in parallel, we cannot use `QueryIterator` in an elegant way like in this solution because it is only sequential. Instead, before iterating, we know on which archetypes the queried entities are on and the size of them. Then, each segment will contain a range of one or multiple archetypes. It should then be possible to fetch while iterating over the entities. Some problems that need to be overcome are that `RwLockReadGuard` is not thread-safe, and that multiple threads need to access disjoint parts of the same vector. I don’t think we need to change anything in this pull request, but maybe create a new issue for this improved implementation.

martinjonsson01:
> To fetch in parallel, we cannot use QueryIterator in an elegant way like in this solution because it is only sequential. Instead, before iterating, we know on which archetypes the queried entities are on and the size of them. Then, each segment will contain a range of one or multiple archetypes. It should then be possible to fetch while iterating over the entities. Some problems that need to be overcome are that RwLockReadGuard is not thread-safe, and that multiple threads need to access disjoint parts of the same vector. I don’t think we need to change anything in this pull request, but maybe create a new issue for this improved implementation.

I'm having a hard time imagining how this would be implemented, perhaps it would be better if you start working on this as your next issue after having finished #50? Unless you've already started working on something else.

EdvinNilsson:
> I'm having a hard time imagining how this would be implemented, perhaps it would be better if you start working on this as your next issue after having finished #50? Unless you've already started working on something else.

I can take a look at this and see if I can find a way of implementing this.

martinjonsson01:
> I can take a look at this and see if I can find a way of implementing this.

Great, thank you, then I'll leave it to you and I'll continue working on the rendering system! 👍

martinjonsson01:
@EdvinNilsson do you think doing it in another branch and merging this one for now is best, or that you continue working on this branch and we wait with merging it until you're also done?

EdvinNilsson:
You can merge this branch.

Issue #87, opened 2023-04-02
Integrate renderer into ECS


martinjonsson01:
## Reason
In order for the renderer to fit well into the rest of the engine, it should be integrated to utilize our ECS.

## Requirements
- [x] The rendering feature is _optional_ and an application can run entirely without it (just like before this feature was implemented) (there should be a feature that enables compilation of this, and it should be possible to compile without)
- [x] When rendering is enabled, rendered frames are shown in the application window
- [x] When rendering is enabled, there is a rendering _system_ which renders entities based off of their visual components
- [x] `fn main` is located in the client code (just like our current example tests), _not_ in the graphics library
- [x] The renderer's update rate (fps) is decoupled from the simulation's update rate (tps), i.e. the simulation can progress at 1000 tps while the renderer is running at 60 fps
- [x] There are components that describe the visual properties of entities:
  - [x] `Mesh` contains information about which mesh an entity should be rendered with
  - [x] `Position` contains information about where an entity is located in world-space
  - [x] `Rotation` contains information about how an entity is rotated in world-space
  - [x] `Scale` contains information about how an entity should be scaled

## Dependencies
- [x] #82

Issue #61, opened 2023-03-21
Rendering system


martinjonsson01:
## Reason
To visualize simulations, render them using [`recs-gfx`](https://github.com/martinjonsson01/gfx).

## Requirements
- [x] #82 
- [x] #86
- [x] #87
- [x] #92

## Dependencies
None

PR #88, opened 2023-04-03
86 application window

### LINKED ISSUE #86: Application window
## Reason
To be able to render the application, it has to have a window in which to place rendered frames.

## Requirements
- [x] When rendering is enabled, an application window is shown 
- [x] `fn main` is located in the client code (just like our current example tests), _not_ in the graphics library
- [x] The rendering feature is _optional_ and an application can run entirely without it (just like before this feature was implemented) (there should be a feature that enables compilation of this, and it should be possible to compile without)
- [x] Rendering of the application window does not prevent nor impede the ECS from running

## Dependencies
- [x] #82

martinjonsson01:
Enables rendering of an application window alongside the ECS running in the background. Part of this meant we had to generalize how plugins are added to the ECS engine. 

> Note: this is a sub-task of #61.

The window is currently completely empty, as rendering of contents will be implemented in #87.
image

The new way of adding a plugin is:
```rust
// Define a trait which describes a property which the application can have:
pub trait Audible {
    fn with_audio(self) -> AudioResult<()>;
}

// Implement the trait for all types of application:
impl<App: Application> Audible for App {
...
}
```
(see `gfx-plugin\lib.rs` for a real example of this)

and now this plugin can be enabled by calling:
```rust
let mut app = BasicApplication::default().with_audio()?;
```

Closes #86


crates/gfx-plugin/examples/rendered-ecs.rs
```diff
@@ -0,0 +1,64 @@
...
+use scheduler::schedule::PrecedenceGraph;
+use std::thread;
+use std::time::Duration;
+use tracing::{instrument, warn};
+
+// a simple example of how to use the crate `ecs`
+#[instrument]
+fn main() -> Result<(), Report> {
+    let mut app = BasicApplication::default()
+        .with_rendering()?
```
EdvinNilsson, line 20: It's nice that it's so easy to add the graphics plugin

crates/gfx-plugin/src/lib.rs
```diff
@@ -0,0 +1,152 @@
+//! A rendering-plugin for RECS, which renders the [`BasicApplication`] using a renderer
```
EdvinNilsson, line 1:
```suggestion
//! A rendering-plugin for RECS, which renders the [`BasicApplication`](ecs::BasicApplication) using a renderer
```

crates/gfx/src/engine.rs
```diff
@@ -100,94 +100,86 @@ pub enum MainMessage {
...
-        + Send
-        + Sync
-        + 'a,
 {
     /// Creates a new instance of `Engine`.
     ///
-    /// The `initialize_gfx` function is called once, before beginning the render loop.
-    /// The `simulate` function is called from the simulation thread every simulation tick.
+    /// [`GraphicsInitializer`] is called once, before beginning the render loop.
+    /// [`Simulation`] is called from the simulation thread every simulation tick.
```
EdvinNilsson, line 138: Cannot find `GraphicsInitializer` and `Simulation` in the code.

crates/ecs/src/profiling.rs
```diff
@@ -19,10 +19,15 @@ pub enum ProfilingError {
 /// Whether a profiling operation succeeded.
 pub type ProfilingResult<T, E = ProfilingError> = Result<T, E>;
 
-impl Application {
+/// Represents an [`Application`] which can be profiled.
+pub trait Profileable: Sized {
     /// Enables profiling connection to Tracy.
+    fn with_profiling(self) -> ProfilingResult<Self>;
```
JacobBredin, line 25: Does this pattern have a name? I'm really enjoying it 😏

EdvinNilsson (approved):
LGTM!

crates/gfx-plugin/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// Initializes and configures rendering functionality, enabling visualization of
+    /// the simulation.
+    fn with_rendering(self) -> GfxAppResult<RenderedApp>;
+}
+
+type GraphicsEngine = gfx::engine::Engine<NoUI, Vec<Object>>;
+type GraphicsEngineHandle = gfx::engine::EngineHandle<Vec<Object>>;
+
+impl<App: Application + Send> Graphical<GraphicalApplication<App>> for App {
+    fn with_rendering(self) -> GfxAppResult<GraphicalApplication<App>> {
```
JacobBredin, line 149: I like the idea of "decorating" the app, makes the builder very clean 👍

JacobBredin (approved):
LGTM!

Cool Ideas used 👍

crates/gfx/src/engine.rs
martinjonsson01: Oops, that's old stuff. We'll remove it.

crates/ecs/src/profiling.rs
martinjonsson01: I'm not sure if it has its own name, but it's inspired by how builders work!

Issue #92, opened 2023-04-04
ECS lights


martinjonsson01:
## Reason
To have control over lighting from the ECS and not have them hardcoded, implement lights in ECS.

## Requirements
- [x] The default rotating light is removed
- [x] There is a component `Light`
- [x] All `Light`-entities are rendered as real lights by the graphics engine

## Dependencies
- [x] #61

PR #93, opened 2023-04-04
87 integrate renderer into ecs

### LINKED ISSUE #87: Integrate renderer into ECS

martinjonsson01:
Entities with specific rendering-components are now rendered in the application window.

In the below gif, the rotating cubes are entirely pure entities in the ECS. The light is not yet converted to ECS (see #92).
Zoom_bm9HabYNEh

The components that make an entity renderable are:
* `Model` which specifies which meshes/materials to use when rendering the entity,
* `Position` which specifies the world-space location of the entity (this needs to be used instead of a custom `Position` component in order for the position of the rendered entity to change),
* `Rotation` which specifies the model-space rotation of the entity,
* `Scale` which can rescale the entity.

These are all currently required on an entity for it to be rendered, but to make construction of visible entities that don't want to specify all of these components simpler, we have created a `RenderedEntityBuilder`:
```rust
let cube_model = app.load_model("cube.obj")?;

let _entity = app
    .rendered_entity_builder(cube_model)?
    .with_position(position)
    .with_scale(scale)
    .build()?;
```
here, the rotation is not specified and instead a default value is used. `Position`, `Scale` and `Rotation` are all optional, but you need to always specify a `Model`.

Closes #87

### DIFF
--- a/crates/gfx-plugin/src/rendering.rs
+++ b/crates/gfx-plugin/src/rendering.rs
@@ -1,0 +1,33 @@
+//! An abstraction over things that are visible in ECS.
+
+pub use cgmath::{Quaternion, Vector3};
+pub use gfx::ModelHandle;
+
+/// A 3D representation of the shape of an object.
+#[derive(Debug)]
+pub struct Model {
+    _handle: ModelHandle,
+}
+
+/// A location in 3D-space.
+#[derive(Debug)]
+pub struct Position {
+    /// The vector-representation of the position.
+    pub vector: Vector3<f32>,
+}
+
+/// An orientation in 3D-space.
+#[derive(Debug)]
+pub struct Rotation {
+    /// Rotation represented as a quaternion.
+    pub rotation: Quaternion<f32>,
+}
+
+/// At which scale to render the object.
+///
+/// Each axis can be scaled differently to stretch the object.
+#[derive(Debug)]
+pub struct Scale {
+    /// The vector-representation of the scale.
+    pub vector: Vector3<f32>,
+}
--- a/crates/gfx-plugin/src/rendering.rs
+++ b/crates/gfx-plugin/src/rendering.rs
@@ -129,7 +129,8 @@ pub struct RenderedEntityBuilder<'app, App> {
 }
 
 impl<InnerApp: Application> GraphicalApplication<InnerApp> {
-    /// Constructs a new [`RenderedEntityBuilder`] which can be used to build
+    /// Constructs a new [`RenderedEntityBuilder`] which can be used to build entities which
+    /// will be visible in the application window.
     pub fn rendered_entity_builder(
         &mut self,
         model: Model,
--- a/crates/gfx-plugin/src/rendering.rs
+++ b/crates/gfx-plugin/src/rendering.rs
@@ -1,33 +1,224 @@
 //! An abstraction over things that are visible in ECS.
-
-pub use cgmath::{Quaternion, Vector3};
+use crate::GraphicalApplication;
+use cgmath::Zero;
+pub use cgmath::{One, Quaternion, Vector3};
+use ecs::systems::{Query, Read};
+use ecs::{Application, Entity};
+use gfx::engine::RingSender;
 pub use gfx::ModelHandle;
+use gfx::Transform;
+use itertools::Itertools;
+use std::error::Error;
+use std::fmt::Debug;
+use thiserror::Error;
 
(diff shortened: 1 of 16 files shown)


crates/gfx-plugin/src/rendering.rs
```diff
@@ -0,0 +1,224 @@
...
+    application: &'app mut App,
+    entity: Entity,
+    model: Model,
+    position: Option<Position>,
+    rotation: Option<Rotation>,
+    scale: Option<Scale>,
+}
+
+impl<InnerApp: Application> GraphicalApplication<InnerApp> {
+    /// Constructs a new [`RenderedEntityBuilder`] which can be used to build
```
EdvinNilsson, line 132: unfinished sentence?

crates/gfx-plugin/src/rendering.rs
```diff
@@ -0,0 +1,224 @@
...
+            position,
+            rotation,
+            scale,
+        } = self;
+
+        application
+            .add_component(entity, model)
+            .map_err(to_component_adding_error)?;
+
+        add_component_or_default(application, position, entity)?;
```
EdvinNilsson, line 193: I've been thinking about whether it would be useful for a `Read`/`Write` system parameter that matches any entity and returns an `Option` with the component. In this case it would be useful so you don't have to add a default component. Do you think adding this system parameter would be a good addition?

EdvinNilsson (approved):
LGTM

crates/gfx-plugin/src/rendering.rs
martinjonsson01: Ohh, so we could have the system query be `rendering_system(..., maybe_rotation: Read<Option<Rotation>>)` and then use a default value if it's `None`? That'd be really nice. 

If it's not a lot of trouble to add then I think it would be a great addition, though I am a bit concerned about how it would affect cache locality. Wouldn't there need to be jumping between several different archetypes to do this? But maybe it's few enough jumps that it's negligible.

martinjonsson01: > Implementing this should be easy, but dynamic archetypes must be implemented first.

Great, would you mind creating an issue for that with the requirements and dependencies needed, then?

JacobBredin (approved):
LGTM!

crates/gfx-plugin/src/rendering.rs
```diff
@@ -0,0 +1,225 @@
...
+
+/// Builds a visible (i.e. rendered) entity.
+#[derive(Debug)]
+pub struct RenderedEntityBuilder<'app, App> {
+    application: &'app mut App,
+    entity: Entity,
+    model: Model,
+    position: Option<Position>,
+    rotation: Option<Rotation>,
+    scale: Option<Scale>,
```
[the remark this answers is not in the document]
martinjonsson01: Yeah, that's an interesting idea. We did consider making it possible to use this builder to add any type of component, but we realized that it could lead to confusion when users try to specify one of the rendering-components using `builder.add_component` rather than `builder.with_position`, because then the default position would overwrite the one specified. 

We tried making a method on `Application` that could add multiple components at the same time, but could not do so in a satisfactory way as a vector of objects needs them to be uniform (i.e. not different component types). You could box all the components but at that point you've already lost so much in ergonomics it's not worth it. But maybe defining prefabs using a trait and a struct could be a way around this.

If you want to you could create an issue for this

PR #97, opened 2023-04-05
Hotfix renderer configuration

martinjonsson01:
The renderer has a bunch of parameters that change how the visuals look, but previously they were not configurable. This PR simply adds builders that make the parameters configurable.

To configure a graphical application, simply do:
```rust
    let mut app = BasicApplicationBuilder::default()
        .with_rendering()?
        .camera_movement_speed(10.0)
        .far_clipping_plane(10_000.0)
        .build()?;
```

### DIFF
--- a/crates/gfx-plugin/src/lib.rs
+++ b/crates/gfx-plugin/src/lib.rs
@@ -37,7 +37,7 @@ pub use cgmath::Deg;
 use crossbeam::channel::Receiver;
 use ecs::systems::{IntoSystem, SystemParameters};
 use ecs::{Application, ApplicationBuilder, Entity, Executor, Schedule};
-use gfx::engine::{Creator, GraphicsOptionsBuilder, RendererOptionsBuilder};
+use gfx::engine::{Creator, GraphicsOptionsBuilder};
 use gfx::engine::{EngineError, MainMessage, NoUI};
 use gfx::time::UpdateRate;
 use std::error::Error;
@@ -52,11 +52,7 @@ use thiserror::Error;
 #[derive(Debug, Default)]
 pub struct GraphicalApplicationBuilder<AppBuilder> {
     app_builder: AppBuilder,
-    camera_movement_speed: Option<f32>,
-    camera_mouse_sensitivity: Option<f32>,
-    far_clipping_plane: Option<f32>,
-    near_clipping_plane: Option<f32>,
-    field_of_view: Option<Deg<f32>>,
+    graphics_options_builder: GraphicsOptionsBuilder,
 }
 
 impl<InnerApp, AppBuilder> ApplicationBuilder for GraphicalApplicationBuilder<AppBuilder>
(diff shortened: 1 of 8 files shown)


crates/gfx-plugin/src/lib.rs
```diff
@@ -50,7 +51,12 @@ use thiserror::Error;
 /// Builds a [`GraphicalApplication`].
 #[derive(Debug, Default)]
 pub struct GraphicalApplicationBuilder<AppBuilder> {
-    builder: AppBuilder,
+    app_builder: AppBuilder,
+    camera_movement_speed: Option<f32>,
+    camera_mouse_sensitivity: Option<f32>,
+    far_clipping_plane: Option<f32>,
+    near_clipping_plane: Option<f32>,
+    field_of_view: Option<Deg<f32>>,
```
Matte02, line 59: I assume these can/will be moved to the `Camera` Component when/if we implement the input system.

crates/gfx-plugin/src/lib.rs
```diff
@@ -74,21 +80,86 @@ where
...
+        self.app_builder = self.app_builder.add_systems(systems);
         self
     }
 
     fn build(self) -> Self::App {
-        let (graphics_engine, graphics_engine_handle) = GraphicsEngine::new()
+        let mut graphics_options_builder = GraphicsOptionsBuilder::default();
+        if let Some(speed) = self.camera_movement_speed {
+            graphics_options_builder.camera_movement_speed(speed);
+        }
```
Matte02, line 91: Is there any reason for why not storing the builder rather than the parameters and allowing the "setters" to change the `graphics_option_builder`? 
Feel like there would be less code over all. The only code you would have here is just the `build` call in 
```rust  
let renderer_options = renderer_options_builder
            .build()
            .expect("default renderer options should be valid");```

File: crates/gfx-plugin/src/lib.rs [hunk 3]
```diff
@@ -74,21 +80,86 @@ where
...
         })
     }
 }
 
+impl<AppBuilder> GraphicalApplicationBuilder<AppBuilder> {
+    /// Sets how fast the camera moves around.
+    pub fn camera_movement_speed(mut self, speed: f32) -> Self {
+        self.camera_movement_speed = Some(speed);
+        self
+    }
```
Matte02, line 130: Following up on the previous comment, here you would only need to do:
```rust
pub fn camera_movement_speed(mut self, speed: f32) -> Self {
        self.graphics_options_builder.camera_movement_speed(speed);
        self
    }
```

Matte02 (approved):
LGTM! Just a small question regarding a design decision.

crates/gfx-plugin/src/lib.rs
martinjonsson01: Ah, that's smart. No reason, I just threw it together extremely quickly. I'll change it to that!

ChristofferPerssons (approved):
Looks good! Can't say I understand everything since I haven't gone thorugh the graphics engine yet but nothing jumped out to me as missing when comparing what changed.

PR #102, opened 2023-04-07
Hotfix scheduler

martinjonsson01:
It turns out the makespan minimization algorithm from #48 that was implemented in #78 is incorrect, and does not work in some cases. Therefore it will be removed, until we can come up with a solution for it.

The most glaring example of where the algorithm breaks down comes from the n-body case study (where systems `gravity` and `acceleration` are incorrectly scheduled to run concurrently), but a simplified example will be given here. If the unoptimized schedule looks like this:
image

then the makespan minimization incorrectly reordered it to be like:
image

The incorrect part is that now both `gravity` and `movement` read and write to `Position` concurrently. 

This PR includes a proptest which generates a large number of random schedules and checks to see if they've incorrectly included any concurrent reads and writes to the same component (we already had a test just like it, but checking for concurrent writes to same component). Since that test fails when makespan minimization is run, and passes when it is disabled, that most likely means our baseline scheduler is still correct.

An attempt at resolving this was done on branch `hotfix-system-reordering`, but it did not work. If anyone has an idea for how to solve this, then please take a look at that branch and continue working on it.


martinjonsson01:
A fix for this was created by Christoffer in #103

PR #104, opened 2023-04-09
89 parallel parameter fetching

### LINKED ISSUE #84: Intra-system segmentation heuristic
## Reason
To avoid too much overhead while still better utilizing the available processor cores, intra-system task segmentation should be done based on a reasonably smart heuristic.

## Requirements
- [x] System segment size is dynamically calculated based on each systems' characteristics
  - [x] (can optionally be replaced with better idea) Segment size is based on how many entities match the query of the system

## Dependencies
- [x] #80

### LINKED ISSUE #89: Parallel parameter fetching
## Reason
In order to speed up iteration of systems where the dominating factor / bottleneck is parameter fetching rather than system computation, parameters should be fetched in parallel.

## Requirements
- [x] Segmented iteration of a `System` also segments fetching of parameters, so it doesn't fetch them all at once sequentially

## Dependencies
- [x] #80

EdvinNilsson:
This pull request adds parallel system execution for both fetching and system execution.

The implementation relies on scoped threads to manage the borrowed data and the `RwLock`s. Because of this, Rayon's global thread-pool is used instead of our `WorkerPool`. To implement this, many changes were made the `SystemParameter` trait to allow multi-threading.

The `Query` system parameter had to be reworked to work with the new `SystemParameter` trait. 
`Query` does no longer implement `IntoIterator`, instead `Query` implements `Iterator`. This change also resolves the previous safety issue where access to component vectors was possible without holding the `RwLock`.

Segment size can also automatically be determined by entity- and thread count.

Closes #89, closes #84


crates/ecs/src/systems/iteration.rs
```diff
@@ -52,163 +58,107 @@ macro_rules! impl_sequentially_iterable_system {
...
+                            });
 
-                    segments
+                        } else if let ($(Some(Some([<parameter_$parameter>])),)*) = (
+                            $([<P$parameter>]::fetch_parameter([<segments_$parameter>].get_mut(0).expect("there should always be at least one segment")),)*
+                        ) {
+                            (self.function)($([<parameter_$parameter>],)*);
+                        }
+                    }
+                    Ok(())
```
martinjonsson01, line 143: This implementation kind of makes `WorkerPool` entirely redundant, since system execution no longer runs there at all (only the dispatching of rayon-tasks is done by the worker threads). The first requirement of #80 has been undone

The reasons I didn't go all-out using Rayon for neither inter-system nor intra-system parallelization is:
* because we have greater control of everything regarding system execution when done in our own implementation of a thread pool. If we want to benchmark/profile/try out something that concerns how systems are executed then we have full control over it and can implement it according to our specific needs. I'm not sure how our shutdown-functionality would be integrated into an implementation entirely using Rayon, for example. The profiler doesn't seem to handle it well either (worker threads doing no work at all show up, so we've got unnecessary threads visible). We can't enable any profiling of the innards of the rayon thread pool either, so we'll have to infer that from the blanks and use other forms of profiling. These examples are solvable of course, but they illustrate the wider point of how it's beneficial to stick with our own implementation.
* because having two separate thread pools causes unnecessary context switching by the operating system needing to multiplex two sets of thread pools on the same logical cores. Since (I don't believe at least) we can't take control over the Rayon threads and make the tasks execute on our own implementation of workers, we should either use Rayon entirely and do away with `WorkerPool`, or continue using `WorkerPool`.
* because a major part of our thesis work is the scheduler, and it seems a bit strange to outsource one of the interesting parts of the ECS to a third-party. There are interesting things we can write about in both results and discussion regarding how the `WorkerPool` doles out tasks and what its downsides/benefits are. We can of course still write about those concepts, but they won't be nearly as relevant as when it's about the actual implementation we've written during the project. 

I'm not saying we shouldn't use Rayon at all, but I think the core parts of the ECS execution should be done using the same thread pools.

Since I never stated these reasons out loud I totally understand that you chose to do it this way, and while I do think that the above reasons are very compelling, I can be convinced that it's worth it to switch to Rayon. There just needs to be a very solid reason to do it, to outweigh the current reasons to keep using our `WorkerPool`.

martinjonsson01: The above are my reasons for why we should stick with `WorkerPool`, but what's more relevant is what your reasons for using Rayon over `WorkerPool` are. 

> The implementation relies on scoped threads to manage the borrowed data and the `RwLock`s. Because of this, Rayon's global thread-pool is used instead of our `WorkerPool`. To implement this, many changes were made the `SystemParameter` trait to allow multi-threading.

I suppose the reason you needed scoped threads was the same as for why I needed scoped threads inside the `WorkerPool` implementation, that is the threads need references to `World` (or to stuff within `World`) and thus need to run for a shorter lifetime than the `World`-reference does (correct me if I'm wrong, maybe you had another reason). What I'm wondering is why this couldn't be integrated to use the already-scoped threads of `WorkerPool`? 

If the problem is that our thread pool isn't globally accessible (so we'd have to be much more careful about flow control and passing it around), then I think that could be remedied quite easily by adding a static variable containing it. 

If the problem is that the segments can't be created in the same function as they're executed (due to them not living long enough), then could we possibly split up the segmentation of a parallel system and the execution of those segments (much like system parameters already do for borrowed data).

martinjonsson01 (changes requested):
Very nice! I got a nice +20 TPS bump from this on n-body.

I think there's a decision that needs to be made here regarding whether to keep using `WorkerPool` or convert it to `Rayon` entirely, which is why I'm requesting changes. See my comment in the `ParallelIterable`-impl inside the macro.

The rest of my comments are mostly nitpicks and a few questions.

crates/ecs/src/systems/mod.rs
```diff
@@ -361,46 +517,96 @@ impl<Component: Debug + Send + Sync + 'static + Sized> SystemParameter for Read<
...
+
+            let chunks = right.chunks_mut(segment_size);
+            for chunk in chunks {
+                if chunk.len() == segment_size {
+                    segments.push((0, vec![(0, chunk)]));
+                } else {
+                    current_segment_size += chunk.len();
+                    current_segment.push((0, chunk));
+                }
+            }
```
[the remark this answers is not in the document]
EdvinNilsson: > Looks like you've got two different ways of handling the remainder (one way in Read and now another one here). Is there a reason for that? Or maybe you just wrote one first and then came up with a better way of doing it later and wrote the other.

The `Write` implementation could not use `chunks_exact_mut` and `into_remainder` because both `into_iter` and `into_remaider` takes the ownership of the iterator. This is why the `Write` implementation is using `chunks_mut` instead and an if expression to check if it is the remainder.

> Either way, the implementations of split_borrowed_data are very similar, could you extract that into a helper-function and use that in both implementations?

The code looks very similar, but most of the types are different and immutable/mutable version of functions are used for the different implementation. This makes it difficult to use helper functions.

crates/ecs/src/systems/mod.rs
```diff
@@ -607,61 +802,110 @@ macro_rules! impl_system_parameter_function {
...
+                            // Borrowed is never mutated except when the Write system parameter is used.
+                            // To ensure safety, intra-parallelization is disabled if any system parameter needs write access.
+                            // This check is done using the "support_parallelization" method on the Query system parameter.
+                            for _ in 0..segment_count {
+                                segments.push((
+                                    ($([<P$parameter>]::split_borrowed_data(unsafe { &mut *[<borrowed_pointer_$parameter>] }, FixedSegment::Single),)*),
+                                    borrowed.1,         // world
+                                    borrowed.2.clone(), // archetypes
+                                ));
+                            }
```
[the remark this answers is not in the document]
EdvinNilsson: > Why do read-only segments need a mutable reference?

All `SystemParameter` use the same trait, and `Write` needs a mutable reference, so all `SystemParameter` will get a mutable reference.

> Why is unsafe being used instead of split_at_mut?

All segments need to reference all components, because this is for foreign entity access. `split_at_mut` is used in `split_borrowed_data` for `Write` instead. 

> Why can't two segments write to different parts of the component vec at the same time? "intra-parallelization is disabled if any system parameter needs write access"

Here, "if any system parameter needs write access" is referring the system parameters that are inside the `Query`. `Query<(Write<A>,)>` cannot use intra-parallelization because all threads will reference all `A` components.

> Is only intra-system parallelization of component fetching disabled if any parameter needs write access, or execution as well?

Both component fetching and execution if any parameter in `Query` needs write access. 
```rust
fn system_a(a: Read<A>, query: Query<(Write<B>,)>); // Will run sequentially
fn system_b(mut a: Write<A>, query: Query<(Read<B>,)>); // Will use intra-system parallelization
```

I will clarify this comment.

crates/ecs/src/systems/mod.rs
```diff
@@ -607,61 +802,110 @@ macro_rules! impl_system_parameter_function {
...
-                        $([<P$parameter>]::borrow(self.world, &archetypes_indices)?,)*
-                    );
+                fn component_accesses() -> Vec<ComponentAccessDescriptor> {
+                    <($([<P$parameter>],)*) as SystemParameters>::component_accesses()
+                }
 
-                    let iterate_over_entities = $([<P$parameter>]::iterates_over_entities() ||)* false;
+                fn iterates_over_entities() -> bool {
+                    false
+                }
```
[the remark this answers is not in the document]
EdvinNilsson: I see the confusion. The `ìterate_over_entites` function tells whether a system needs to iterate over the entities when fetching parameters. See this example:

```rust
fn system_a(a: Read<A>, query: Query<(Read<A>,)>);
fn system_b(query: Query<(Read<A>,)>);
```

When running `system_a`, it will iterate over all entities with component `A`.
But with `system_b`, this system will not iterate and only run once per tick.

Do you have any suggestion on what `ìterate_over_entites` could be called instead?

Also, `$([<P$parameter>]::iterates_over_entities() ||)* false` is also completely unrelated: This function was not changed, it was just moved (previously on line number 530).

crates/ecs/src/systems/mod.rs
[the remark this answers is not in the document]
martinjonsson01: Alright, fair enough 👍

crates/ecs/src/systems/mod.rs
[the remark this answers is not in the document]
martinjonsson01: Ah, I see. It seems I had missed that this implementation was specifically for `Query` and thus focused on foreign entity access. I think I imagined that we still use the `Query` for normal system execution as well.

> I will clarify this comment.

Thanks!

crates/ecs/src/systems/mod.rs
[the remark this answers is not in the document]
martinjonsson01: > Also, `$([<P$parameter>]::iterates_over_entities() ||)* false` is also completely unrelated: This function was not changed, it was just moved (previously on line number 530).

Oh, that was confusing lol

> The `ìterate_over_entities function tells whether a system needs to iterate over the entities when fetching parameters.

Why specifically when fetching parameters? In the example, it seems like both `system_a` and `system_b` would need to fetch parameters for multiple entities, the difference between them being how many times the system itself is called. 

> Do you have any suggestion on what `iterates_over_entities` could be called instead?

Maybe `contains_multiple_entities`, since it's a system parameter which contains data from multiple entities while the other system parameters only contain data from a single entity. (But you'd have to negate the implementations and uses of it, in this case.) I think that would make it more understandable, since it more clearly describes what aspect of the system parameter that it is giving information about, rather than instructing an executor about how to iterate over it.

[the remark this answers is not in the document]
EdvinNilsson: > Why specifically when fetching parameters? In the example, it seems like both system_a and system_b would need to fetch parameters for multiple entities, the difference between them being how many times the system itself is called.

`fetch_parameter` for `Read`/`Write` will return `Some(Component)` and then `None` when it reaches the end of the last archetype.
`fetch_parameter` for `Query` will always return `Some(Query)` and the actual parameter fetching happens in `fetch_parameter` for the parameters inside `Query` when iterating over it.
If `fetch_parameter` eventually will return `None`, then `ìterate_over_entites` is true.

> Maybe contains_multiple_entities, since it's a system parameter which contains data from multiple entities while the other system parameters only contain data from a single entity.

`contains_multiple_entities` would only make sense for `Query`, `iterate_over_entites` is also false for filters and will be false for resources and command buffers.

crates/ecs/src/systems/iteration.rs
EdvinNilsson: I completely agree with your points. For an implementation like this to work, scoped threads (or something similar) are needed and the easiest way for me to implement this was to use rayon. I know that this is not optimal. You can see this implementation more like a proof of concept. I think that we should try to see if it is possible to integrate this with our `WorkerPool`, but I do not know the `WorkerPool` implementation that well. Maybe you could take a look at this?

> I suppose the reason you needed scoped threads was the same as for why I needed scoped threads inside the WorkerPool implementation, that is the threads need references to World (or to stuff within World) and thus need to run for a shorter lifetime than the World-reference does (correct me if I'm wrong, maybe you had another reason). What I'm wondering is why this couldn't be integrated to use the already-scoped threads of WorkerPool?

Yes, the threads need to run for a shorter lifetime than the `World` reference and the `borrowed` reference. Here are some more details about the constraints:
* `borrowed` cannot outlive the `World` reference.
* `borrowed` does not implement `Send` because of the `RwLockGuard`.
* `segment` is moved into the thread, `segment` cannot outlive `borrowed`.
* `borrowed` needs to be dropped when all threads are done executing, so the `RwLock`s can be released.

> If the problem is that our thread pool isn't globally accessible (so we'd have to be much more careful about flow control and passing it around), then I think that could be remedied quite easily by adding a static variable containing it.

This is not a problem.

> If the problem is that the segments can't be created in the same function as they're executed (due to them not living long enough), then could we possibly split up the segmentation of a parallel system and the execution of those segments (much like system parameters already do for borrowed data).

I don't know if I get your point, but one task needs to borrow the data, create all segments and then fork the segments to different tasks. Then join the tasks and drop the borrowed data.

Do you think this could be implemented to the `WorkerPool`?

crates/ecs/src/systems/mod.rs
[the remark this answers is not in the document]
martinjonsson01: > `fetch_parameter` for `Query` will always return `Some(Query)` and the actual parameter fetching happens in `fetch_parameter` for the parameters inside `Query` when iterating over it.

Ah, I see. So  `Read`/`Write` parameters are eagerly fetched, while `Query` parameters are lazily fetched?

> If `fetch_parameter` eventually will return `None`, then `iterates_over_entities` is true.

> `contains_multiple_entities` would only make sense for `Query`, `iterates_over_entities` is also false for filters and will be false for resources and command buffers.

Then it sounds like the function represents whether the parameter contains a finite amount of data. Ordinary `Read`/`Write` parameters only contain data once for each entity, which is why they eventually return `None`, while special parameters such as `Query` and filters always return something and is there for not finite. 

So maybe it could be renamed to `contains_finite_data` or similar? Or maybe that's even more vague, heh. What do you think?

(As an aside: the whole `None`/`Some`-business and pseudo-iterator in `Read` and `Write` is still not ideal. Now that you seem to have figured out a way to implement `Iterator` for `Query`, do you think we could make all `SystemParameter`s into iterators?)

crates/ecs/src/systems/iteration.rs
martinjonsson01: > I completely agree with your points. For an implementation like this to work, scoped threads (or something similar) are needed and the easiest way for me to implement this was to use rayon. I know that this is not optimal. You can see this implementation more like a proof of concept. 

I see. I'm really glad we're on the same page regarding this! 

> * `borrowed` does not implement `Send` because of the `RwLockGuard`.
> * `segment` is moved into the thread, `segment` cannot outlive `borrowed`.

IIRC `segment` contains references to `borrowed`, but like you say `borrowed` is not `Send`. Then how can `segment` be `Send`?

>> If the problem is that the segments can't be created in the same function as they're executed (due to them not living long enough), then could we possibly split up the segmentation of a parallel system and the execution of those segments (much like system parameters already do for borrowed data).

>I don't know if I get your point, but one task needs to borrow the data, create all segments and then fork the segments to different tasks. Then join the tasks and drop the borrowed data.

Hm, I think I understand. 

> I think that we should try to see if it is possible to integrate this with our `WorkerPool`, but I do not know the `WorkerPool` implementation that well. Maybe you could take a look at this?

Sure thing! I'll give it a try. Though I suspect we might need to do this together, as the feeling is mutual and I don't know the `SystemParameter` implementation that well 😝 

Do you think I should create a branch off of `89-parallel-parameter-fetching` and try to integrate it into our `WorkerPool`, or just do it directly in here?

EdvinNilsson: > IIRC segment contains references to borrowed, but like you say borrowed is not Send. Then how can segment be Send?

`segment` stores a slice of the component vec in `borrowed` that is `Send`. It is just the `RwLockGuard` that is not `Send` because it is required to be dropped on the same thread as it is created on.

> Do you think I should create a branch off of 89-parallel-parameter-fetching and try to integrate it into our WorkerPool, or just do it directly in here?

I think it is better to create a new branch from this, so this branch remains in functional state to test other things.

martinjonsson01: Alright, I'll make a new branch and explore a bit. In the mean time, you can finish up any remaining comments in this PR and then move on to other things, as if this PR is completed. I'll let you know if (or maybe when) I get stuck and we need to do the rest together. 

That sound good?

ChristofferPerssons (approved):
LGTM! You've already discussed using our Workepool instead of Rayon which I think would be preferable if possible. 
I admit that I sometimes have trouble following the macros. Are there any ways to get a preview that removes the "macroness"?

EdvinNilsson:
> Are there any ways to get a preview that removes the "macroness"?

You can expand the macro in CLion using the context actions menu from the line where the macro is used.

crates/ecs/src/systems/mod.rs
```diff
@@ -596,7 +595,18 @@ impl<'components, Component: Debug + Send + Sync + 'static + Sized> Clone
...
-        panic!("Cloning a segment for a mutable SystemParameter is not allowed.")
+        Self {
+            slices: self
+                .slices
+                .iter()
+                .map(|slice| {
+                    let ptr = (*slice).as_ptr();
+                    let len = slice.len();
+                    // SAFETY: Write segments will only be cloned when intra-parallelization is disabled.
+                    unsafe { slice::from_raw_parts_mut(mem::transmute(ptr), len) }
```
martinjonsson01, line 606: Is there somewhere we could add some assertions (or `debug_assert`s if they impact performance) that checks that this is maintained?

crates/ecs/src/systems/mod.rs
```diff
@@ -710,14 +738,67 @@ impl<'a, Component: Debug + Send + Sync + 'static + Sized> SystemParameter
     }
 }
 
-/*
+/// Segment for [`SystemParameter`] without data.
+#[derive(Debug, Default)]
+pub struct UnitSegment<T: Default> {
+    item: PhantomData<T>,
```
martinjonsson01, line 744: What's the need for `T` when it's "without data", why not make it a unit struct?

crates/ecs/src/systems/mod.rs
```diff
@@ -710,14 +738,67 @@ impl<'a, Component: Debug + Send + Sync + 'static + Sized> SystemParameter
...
+    slices: Vec<&'components [Entity]>,
+}
+
+impl<'components> Clone for EntitySegment<'components> {
+    fn clone(&self) -> Self {
+        Self {
+            slices: self.slices.clone(),
+        }
+    }
+}
```
martinjonsson01, line 787: How come this is necessary, instead of deriving it? Since both `Vec<T: Clone>: Clone` and `&T: Clone`, I'd imagine that it's possible to derive, but I might be missing something

martinjonsson01: Same goes for `UnitSegment`

crates/ecs/src/systems/mod.rs
```diff
@@ -777,51 +858,48 @@ impl SystemParameter for Entity {
...
-                *component_index += 1;
-                Some(entity)
-            } else {
-                // End of archetype
-                *current_archetype += 1;
-                Self::fetch_parameter(borrowed)
-            };
-        }
-        // No more entities
-        None
```
martinjonsson01: 🤩

crates/ecs/src/systems/mod.rs
```diff
@@ -1018,49 +1098,52 @@ macro_rules! impl_system_parameter_function {
...
+            impl<'a, $([<P$parameter>]: SystemParameter + 'a,)*> IntoIterator for QuerySegment<'a, ($([<P$parameter>],)*)> {
                 type Item = Query<'a, ($([<P$parameter>],)*)>;
                 type IntoIter = impl Iterator<Item=Self::Item>;
 
                 fn into_iter(mut self) -> Self::IntoIter {
                     iter::from_fn(move || {
                         Some(Query {
+                            // SAFETY: All Query instances need a mutable reference the segment for foreign entity access.
+                            // The components are only mutated when Write is a nested system parameter in Query.
+                            // In this case, intra-system-parallelization is disabled to maintain mutual exclusion.
```
martinjonsson01, line 1125: Same here, is there a fitting spot where we could add assertions that check this is always true? I.e. `assert!(query_has_no_write_parameters || !query_supports_intra_system_parallelization)`

martinjonsson01 (approved):
Very nice!

Feels great to finally be rid of those horrendous `fetch_parameter` implementations!

martinjonsson01:
Also, this error in the compiler is a bit worrying:
```
error: internal compiler error: compiler/rustc_monomorphize/src/collector.rs:1038:9: no MIR available for DefId(30:2986 ~ wgpu_core[8ff8]::track::texture::{impl#0}::all_ordered)
```

Although it builds fine on my machine, it's only the GitHub runner failing. Maybe because it's using the latest nightly compiler, hmm... If that's the case then we might have to lock the project to a specific version of nightly.

.idea/recs.iml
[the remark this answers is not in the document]
martinjonsson01: Yeah the project isn't loading correctly for me now when I pull it down. Could you restore this file?

crates/ecs/src/systems/mod.rs
EdvinNilsson: A segment needs to implement `IntoIterator<Item=Self>` where `Self` is the system parameter. `T` is the system parameter that the segment is for.

crates/ecs/src/systems/mod.rs
EdvinNilsson: The code generated with derive usually has unnecessary strict trait bounds when working with slices and `PhantomData`. This is why `ReadSegment` cannot derive `Clone`, filters cannot derive `Default` and `UnitSegment` cannot derive `Clone`. `EntitySegment` is the exception because `Entity` implements `Clone`, so in this case it would work.

crates/ecs/src/systems/mod.rs
EdvinNilsson: I don’t think there is a good way of checking this. The way you can see if a parameter has write access is to use component accesses and this is exactly what `supports_parallelization` does, so it would essentially be an asset for `supports_parallelization == supports_parallelization`.

martinjonsson01: Oh, well. Then let's not!

EdvinNilsson:
I tried nightly (2023-05-06) on my machine and it also fails, but nightly (2023-05-05) works.

martinjonsson01:
Then let's [pin the project](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) to nightly 2023-05-04.

(I'm not sure if the Rust-installation in the CI scripts will recognize that file, so if they don't you may have to change them manually)

JacobBredin (approved):
Fantastic, LGTM! Nice to see the performance gains!

I tried to find a way to avoid some of the cloning done, but I leave empty handed. I assume this is what is left in terms of obvious optimizations?

PR #107, opened 2023-04-17
Hotfix Remove redundant option from ComponentVec

JacobBredin:
The extra Option wrapper has been removed from ComponentVec and its associated types. All tests pass, the small example programs runs. The n-body simulation runs and profiling still works. 

Please also run these on your system as well, so that we know it really works. Thanks 😄

### DIFF
--- a/crates/ecs/src/filter.rs
+++ b/crates/ecs/src/filter.rs
@@ -39,10 +39,10 @@ impl<Component: Debug + Send + Sync + 'static + Sized> SystemParameter for With<
         Ok(())
     }
 
-    unsafe fn fetch_parameter(_: &mut Self::BorrowedData<'_>) -> Option<Option<Self>> {
-        Some(Some(Self {
+    unsafe fn fetch_parameter(_: &mut Self::BorrowedData<'_>) -> Option<Self> {
+        Some(Self {
             phantom: PhantomData::default(),
-        }))
+        })
     }
 
     fn component_accesses() -> Vec<ComponentAccessDescriptor> {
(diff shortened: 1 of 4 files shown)


martinjonsson01 (approved):
Looks ___SO___ good to me! 🤩

martinjonsson01:
Though, I think what would give the actual performance gains and what's more meaningful is to remove the redundant `Option<..>` here:
https://github.com/martinjonsson01/recs/blob/ac1555d07ca6dbd247dc1a4a9fb3accf56a6ada4/crates/ecs/src/lib.rs#L1041

Could you take a look at that as well? Removing this `Option` would probably make iteration over the component vecs quicker. (Removing `Option` would remove the enum discriminant meaning we could free up several bytes per component!)

JacobBredin:
> Though, I think what would give the actual performance gains and what's more meaningful is to remove the redundant `Option<..>` here:
> 
> https://github.com/martinjonsson01/recs/blob/ac1555d07ca6dbd247dc1a4a9fb3accf56a6ada4/crates/ecs/src/lib.rs#L1041
> 
> Could you take a look at that as well? Removing this `Option` would probably make iteration over the component vecs quicker. (Removing `Option` would remove the enum discriminant meaning we could free up several bytes per component!)

Yep. First time I did the refactor I failed, so I started over and must have forgot to remove them.😑 

Good catch! I'm not so sure what I removed then.

Matte02 (approved):
LGTM 👍

martinjonsson01 (approved):
Beautiful!

Matte02 (approved):
👍 👍 👍

PR #108, opened 2023-04-19
66 allow systems to access the iterated entity

linnosterlund:
Added a function that takes in the component index and returns the entity that has said component. Will be useful when adding/removing components from entities. 

```rust
fn entity_system(a: Read<A>, entity: Entity){
    info!("executing entity_system");
}
```


Matte02 (approved):
LGTM!

crates/ecs/src/lib.rs
```diff
@@ -375,7 +375,7 @@ pub type ArchetypeResult<T, E = ArchetypeError> = Result<T, E>;
 
 /// Stores components associated with entity ids.
 #[derive(Debug, Default)]
-struct Archetype {
+pub struct Archetype {
```
martinjonsson01, line 378: Why `pub`?

linnosterlund: To reach Archetype in systems/mod.rs, the system parameter Entity needs to save the archetypes it has borrowed to find the entities.

martinjonsson01: I see, then it's preferable to make it crate-private!

```suggestion
pub(crate) struct Archetype {
```

linnosterlund: Had to revert since there was one more thing that I forgot: Archetype is used in Entity's interface which is public and therefore needs to be public. Is it ok that Archetype is public or do we want to avoid it?

martinjonsson01: It's not a dealbreaker, just something that'd be nice to avoid (ideally, archetypes are an implementation-detail, entirely hidden from outside view). If it's not possible to solve by making the other stuff private as well, then you can just leave it as-is.

linnosterlund: it is here that it is public: 
```rust
impl SystemParameter for Entity {
    type BorrowedData<'archetypes> = (
        BorrowedArchetypeIndex,
        Vec<(ComponentIndex,&'archetypes Archetype)>,
    );
```

Basically, the systam parameter saves the archetypes in a vector when it does borrow so that it can iterate through them when looking for entity. I can look into it, but I think it's needen.

linnosterlund: I thought I did? Hmm

martinjonsson01: try running `pre-commit run --all-files` and see if any files are modified

PR #111, opened 2023-04-19
Add remove component optimization

ChristofferPerssons:
Should be some performance gains for addition and removal of components. Please test on your own system since the benchmarks have been fluctuating wildly on mine (+-30%).


crates/ecs/src/lib.rs
```diff
@@ -460,8 +464,9 @@ pub type ArchetypeResult<T, E = ArchetypeError> = Result<T, E>;
 /// Stores components associated with entity ids.
 #[derive(Debug, Default)]
 struct Archetype {
-    component_typeid_to_component_vec: HashMap<TypeId, Box<dyn ComponentVec>>,
-    entity_to_component_index: HashMap<Entity, ComponentIndex>,
+    component_typeid_to_component_vec: FnvHashMap<TypeId, Box<dyn ComponentVec>>,
+    entity_to_component_index:
+        HashMap<Entity, ComponentIndex, BuildHasherDefault<NoHashHasher<Entity>>>,
```
martinjonsson01, line 469: Could you type alias this `HashMap` into something like `NoHashHashMap<K, V>` as well?

crates/ecs/src/lib.rs
```diff
@@ -1193,9 +1207,11 @@ impl<T: Debug + Send + Sync + 'static> ComponentVec for ComponentVecImpl<T> {
 #[derive(Default, Debug, Ord, PartialOrd, Eq, PartialEq, Copy, Clone, Hash)]
 pub struct Entity {
     id: usize,
-    _generation: usize,
+    //_generation: usize,
 }
 
+impl IsEnabled for Entity {}
```
martinjonsson01, line 1213: The recommended way to implement `IsEnabled` is to manually implement `Hash`, to ensure the invariant of 
> **One of the Hasher::write_{u8,u16,u32,u64,usize,i8,i16,i32,i64,isize} methods is invoked exactly once.**

is always upheld.

Could you replace our `#[derive(Hash)]` for `Entity` with a manual `Hash`-implementation, like the one shown in [the documentation](https://docs.rs/nohash-hasher/latest/nohash_hasher/trait.IsEnabled.html)?

crates/ecs/src/lib.rs
```diff
@@ -616,13 +622,18 @@ pub struct World {
     /// This will result in two vectors containing the indices of the `Archetype`s that store these types.
     /// By taking the intersection of these vectors you will know which `Archetype` contain both A and B.
     /// This could be the archetypes: (A,B), (A,B,C), (A,B,...) etc.
-    component_typeid_to_archetype_indices: HashMap<TypeId, HashSet<ArchetypeIndex>>,
+    component_typeid_to_archetype_indices:
+        FnvHashMap<TypeId, HashSet<ArchetypeIndex, BuildHasherDefault<NoHashHasher<usize>>>>,
```
martinjonsson01, line 626: How come you're using both `FnvHashMap` and `BuildHasherDefault<NoHashHasher<usize>>`? To me it seems like `FnvHashMap` is redundant when you're using `NoHashHasher`, but maybe I'm missing something.

crates/ecs/src/lib.rs
```diff
@@ -810,6 +821,11 @@ impl World {
                 let target_archetype_idx =
                     self.move_entity_components_to_new_archetype(entity, source_type_ids)?;
 
+                let mut target_type_ids_vec: Vec<TypeId> = target_type_ids.into_iter().collect();
+                target_type_ids_vec.sort();
```
martinjonsson01, line 825: You should leave a comment describing why these need to be sorted in order for it to work.

crates/ecs/src/lib.rs
```diff
@@ -861,7 +877,14 @@ impl World {
                 )?;
             }
             None => {
-                self.move_entity_components_to_new_archetype(entity, target_type_ids)?;
+                let mut target_type_ids_vec: Vec<TypeId> =
+                    target_type_ids.clone().into_iter().collect();
+                let target_archetype_idx =
+                    self.move_entity_components_to_new_archetype(entity, target_type_ids)?;
+
+                target_type_ids_vec.sort();
```
martinjonsson01, line 885: here too (like my other comment)

crates/ecs/src/lib.rs
```diff
@@ -1193,9 +1207,11 @@ impl<T: Debug + Send + Sync + 'static> ComponentVec for ComponentVecImpl<T> {
 #[derive(Default, Debug, Ord, PartialOrd, Eq, PartialEq, Copy, Clone, Hash)]
 pub struct Entity {
     id: usize,
-    _generation: usize,
+    //_generation: usize,
```
martinjonsson01, line 1210: This can just be removed, since we never wound up using it at all.

crates/ecs/src/lib.rs
```diff
@@ -963,24 +987,11 @@ impl World {
...
-            let archetype = self
-                .archetypes
-                .get(arch_id)
-                .expect("Archetype should exist since index was just fetched.");
-
-            archetype.component_typeid_to_component_vec.len() == component_type_ids.len()
-        };
-        arch_ids.into_iter().find(contains_matching_components)
+        let mut sorted_component_type_ids = component_type_ids.clone();
+        sorted_component_type_ids.sort();
```
martinjonsson01, line 991: here too (like my other comment)

crates/ecs/src/lib.rs
```diff
@@ -1290,6 +1306,22 @@ mod tests {
 
     // Archetype tests:
 
+    #[test]
+    fn type_id_order_does_not_affects_fetching_of_correct_archetype() {
```
martinjonsson01, line 1310: nit:
```suggestion
    fn type_id_order_does_not_affect_fetching_of_correct_archetype() {
```

crates/ecs/src/systems/mod.rs
```diff
@@ -270,7 +271,10 @@ pub trait SystemParameter: Send + Sync + Sized {
 
     /// Perform a filter operation on a set of archetype indices.
     /// The `universe` is a set with all archetype indices used by the `base_signature`.
-    fn filter(universe: &HashSet<ArchetypeIndex>, world: &World) -> HashSet<ArchetypeIndex> {
+    fn filter(
+        universe: &HashSet<ArchetypeIndex, BuildHasherDefault<NoHashHasher<usize>>>,
+        world: &World,
+    ) -> HashSet<ArchetypeIndex, BuildHasherDefault<NoHashHasher<usize>>> {
```
martinjonsson01, line 277: here you should use the type aliases

martinjonsson01 (changes requested):
LGTM!

I'd just like you to add some type aliases to make everything readable again, and also manually implement `Hash` for `Entity` (see comment for as to why).

crates/ecs/src/lib.rs
ChristofferPerssons: The `NoHashHasher` is faster and prefferable however `FnvHashMap` supports `TypeId` while the identity-hasher does not.  Since `TypeId` is from my understanding, in current rust, an `u64` we could try to use this however it seems like relying on `TypeId` always being an `u64` is discouraged by the rust community.

martinjonsson01: Oh, I see. I've misread this line of code. I first interpreted it as a `FnvHashMap` with its hash-function set to `NoHashHasher` which is what seemed redundant. But now I see that it's a `FnvHashMap` with its value being a `HashSet` using `NoHashHasher`. Once you've implemented the type aliases this should become much clearer to readers (and me lol)

martinjonsson01 (approved):
LGTM!

crates/ecs/src/lib.rs
```diff
@@ -861,7 +893,17 @@ impl World {
                 )?;
             }
             None => {
-                self.move_entity_components_to_new_archetype(entity, target_type_ids)?;
+                let mut target_type_ids_vec: Vec<TypeId> =
+                    target_type_ids.clone().into_iter().collect();
```
EdvinNilsson, line 897: Use `.iter().cloned()` instead of `.clone().into_iter()` to reduce unnecessary allocation. Same thing on line number 716.

crates/ecs/src/lib.rs
```diff
@@ -457,11 +461,17 @@ pub enum ArchetypeError {
...
+
 /// Stores components associated with entity ids.
 #[derive(Debug, Default)]
 struct Archetype {
-    component_typeid_to_component_vec: HashMap<TypeId, Box<dyn ComponentVec>>,
-    entity_to_component_index: HashMap<Entity, ComponentIndex>,
+    component_typeid_to_component_vec: FnvHashMap<TypeId, Box<dyn ComponentVec>>,
+    entity_to_component_index: NoHashHashMap<Entity, ComponentIndex>,
     entity_order: Vec<Entity>,
 }
```
EdvinNilsson, line 476: The function `store_entity` can be replaced with this to to get rid of the unnecessary call to `entity_to_component_index.contains_key(&entity)`.
```rust
fn store_entity(&mut self, entity: Entity) -> ArchetypeResult<()> {
    let entity_index = self.entity_to_component_index.len();

    match self.entity_to_component_index.insert(entity, entity_index) {
        None => {
            self.entity_order.push(entity);
            Ok(())
        }
        Some(_) => {
            Err(ArchetypeError::EntityAlreadyExists(entity))
        }
    }
}
```

EdvinNilsson (approved):
LGTM!

martinjonsson01:
@ChristofferPerssons we should wait with merging this until #109 is merged!

PR #115, opened 2023-04-26
55 benchmark n body system in bevy vs ours

martinjonsson01:
Benchmarks the n-body case study in both Bevy and RECS.

The implementation required a refactor of the n-body implementation so that it's extracted into a lib-crate. Therefore the demo-app we previously had is now an example of how to use the n-body lib.

The benchmarks simulate at least `100` ticks and then calculate how much time each tick takes on average. In the below graph, `2^n` bodies with n in range `[0, 14]` are benchmarked.

image


crates/n-body/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+pub fn acceleration_due_to_gravity(
+    position: &Position,
+    acceleration: &mut Acceleration,
+    bodies_query: impl Iterator<Item = (impl Deref<Target = Position>, impl Deref<Target = Mass>)>,
+) {
+    let Position(rendering::Position { point: position }) = position;
+    let Acceleration(ref mut acceleration) = acceleration;
+
+    let acceleration_towards_body = |(body_position, body_mass): (Point3<f32>, f64)| {
+        let to_body: Vector3<f64> = (body_position - position)
```
EdvinNilsson, line 180: Can we also replace f64 with f32.

EdvinNilsson (approved):
In general, the code looks good, but I have a few questions about the implementation.

This benchmark uses a different method from the way the other n-body benchmarks are implemented, making them less comparable. I think this could be a problem when describing the method in the thesis. What do you think?

I see that you have implemented the systems in a way where you share as much code as possible between Bevy and RECS. I would not have done it in this way because you cannot use the native components for Position/Transform directly, you are not using the vector library (glam) that Bevy generally is used with, and you also lack an easy side-by-side code comparison between the engines.

martinjonsson01:
> This benchmark uses a different method from the way the other n-body benchmarks are implemented, making them less comparable. I think this could be a problem when describing the method in the thesis. What do you think?

That's true. I think the way it's being done here is a lot cleaner and more encompassing, since the benchmark is quite literally executing one tick at a time. But I can see how this approach is more difficult in Unity since you can't manually pump the update-loop there (afaik). I'm not sure how these benchmarks would be implemented from within a system when running in Criterion through, since it benchmarks the total runtime of a function call. So I'm not sure if that's possible, do you have any ideas for how that could be done?

> you have implemented the systems in a way where you share as much code as possible between Bevy and RECS. I would not have done it in this way because you cannot use the native components for Position/Transform directly

The main benefit of comparing directly against Rust-native ECS:es is that we're able to use the exact same code, ensuring total implementation fairness. Now, you could argue that this common implementation is currently biased towards RECS because it uses the `Position`-component from `gfx-plugin`, which it sounds like you are. I really don't see how this is to our benefit, since the only reason this component comes from `gfx-plugin` and is not entirely defined within the common n-body implementation is so it can be queried by the rendering system of RECS. 

> you are not using the vector library (glam) that Bevy generally is used with

Why would using native components and Bevy's vector library make it a better comparison? Right now we're using *exactly the same components and vector library* which to me is a lot fairer of a comparison. Otherwise we're introducing the confounding factor of how differently they've implemented their vector math to the `cgmath` library we're currently using in the common implementation. By using the same math library, what we're benchmarking is entirely the difference in ECSes, not the difference in vector libraries.

> you also lack an easy side-by-side code comparison between the engines

Yeah, this is a fair point. I considered this when I first refactored out the common implementation, since we're now losing the very clear example of "this is what a self-contained RECS-app looks like". I did it anyways because the main point of the case study is not the visuals of how the implementations look but rather how they perform. The visual/practical implementation differences are better considered in isolation rather than on a whole case study imo.

EdvinNilsson:
> > This benchmark uses a different method from the way the other n-body benchmarks are implemented, making them less comparable. I think this could be a problem when describing the method in the thesis. What do you think?
> 
> That's true. I think the way it's being done here is a lot cleaner and more encompassing, since the benchmark is quite literally executing one tick at a time. But I can see how this approach is more difficult in Unity since you can't manually pump the update-loop there (afaik). I'm not sure how these benchmarks would be implemented from within a system when running in Criterion through, since it benchmarks the total runtime of a function call. So I'm not sure if that's possible, do you have any ideas for how that could be done?

I agree that Criterion is making it more elegant, but as you say, an implementation like this would not be possible in Unity for example. On the other hand, using an average across multiple ticks using the engine's normal update loop could also be more accurate. If this is hard to implement using Criterion, maybe we could just leave the Criterion implementation as is and use another method when comparing with the other non-rust engines.



> > you have implemented the systems in a way where you share as much code as possible between Bevy and RECS. I would not have done it in this way because you cannot use the native components for Position/Transform directly
> 
> The main benefit of comparing directly against Rust-native ECS:es is that we're able to use the exact same code, ensuring total implementation fairness. Now, you could argue that this common implementation is currently biased towards RECS because it uses the `Position`-component from `gfx-plugin`, which it sounds like you are. I really don't see how this is to our benefit, since the only reason this component comes from `gfx-plugin` and is not entirely defined within the common n-body implementation is so it can be queried by the rendering system of RECS.
> 
> > you are not using the vector library (glam) that Bevy generally is used with
> 
> Why would using native components and Bevy's vector library make it a better comparison? Right now we're using _exactly the same components and vector library_ which to me is a lot fairer of a comparison. Otherwise we're introducing the confounding factor of how differently they've implemented their vector math to the `cgmath` library we're currently using in the common implementation. By using the same math library, what we're benchmarking is entirely the difference in ECSes, not the difference in vector libraries.

If we could use the same components and vector library across all the benchmarks would make it more fair, but this is not possible when we are comparing againsts Rust, C++ and C# based engines. So the implementations will depend on the compiler, the runtime and the external libraries that are being used, and we need to accept that. So the best thing we can do to make it as fair as possible is to show how the “intended” or most common use-cases perform. For vector libraries, Unity GamesObjects is using “UnityEngine.Vector3”, Unity ECS is using “Unity.Mathematics.float3”, and Bevy is using “glam.Vec3”. Using another vector library may decrease the performance if the engine is optimized for a specific library. For example, the Unity Burst Compiler is optimized for Unity.Mathematics. This is one downside with this case-study is that it is so compute-heavy, so the performance is largely affected by how well the linear algebra is optimized instead of the ECS implementation. That Unity and Bevy is storing all transform data in a single component is also a design-choice of the engines, and this will affect how the engine is generally used and how well it performs.

Matte02 (approved):
LGTM 👍

martinjonsson01:
> using an average across multiple ticks using the engine's normal update loop could also be more accurate

Depending on what you mean by accurate, I suppose. By starting/stopping a timer from within a system you don't measure any setup on the first iteration, or any cleanup on the last. On the other hand, you do measure any possible update-loop overhead within the engine. 

> If this is hard to implement using Criterion, maybe we could just leave the Criterion implementation as is and use another method when comparing with the other non-rust engines

I mean, it's really only Unity hindering us from making the benchmarks easier to work with. In EnTT it'd be very simple to benchmark like Bevy and RECS. But if there's no getting around that then I guess we'll have to do it using benchmark-systems as well. I think it might be difficult to implement repeated sampling of the normal update-loop in RECS, since it'd require multiple threads signaling when to shut down / collect data / reboot and so on... I'll give it a go and see.

> [...] So the best thing we can do to make it as fair as possible is to show how the “intended” or most common use-cases perform. [...]

That makes sense, although I'm not a fan of the reasoning that since we can't do everything fair we should do everything unfair. But it's a good point that it's more realistic to compare them completely in their own ecosystems, so I've implemented that now.

Bevy is now running with its own components using `Vec3` from `glam` and `Transform` instead of `Position`. RECS is now running using `f32` instead of `f64`. The results are:
image

Which is quite unfortunate, since Bevy is about 25% faster at all body counts and RECS is 40% slower at all body counts. The slowdown in RECS must be because of using floats, which is very strange since you'd expect more of them to fit into cache/registers. Maybe my CPU only has 64-bit FPUs... 

With the reasoning that we should use all engines in their own ecosystems, I suppose switching back to doubles for RECS would be okay if those perform better in the engine. But maybe you could try running the bench on your machine as well, to see if floats perform better than doubles for you?

EdvinNilsson:
> Depending on what you mean by accurate, I suppose. By starting/stopping a timer from within a system you don't measure any setup on the first iteration, or any cleanup on the last. On the other hand, you do measure any possible update-loop overhead within the engine.

By starting and stopping may possibly for example avoid some memory bandwidth issues or issues with a full write-buffer that could occur if the update loop where not interrupted.




> I mean, it's really only Unity hindering us from making the benchmarks easier to work with. In EnTT it'd be very simple to benchmark like Bevy and RECS. But if there's no getting around that then I guess we'll have to do it using benchmark-systems as well. I think it might be difficult to implement repeated sampling of the normal update-loop in RECS, since it'd require multiple threads signaling when to shut down / collect data / reboot and so on... I'll give it a go and see.

This is what I have been using then benchmarking RECS. It is a very hacky solution and I have to manually shutdown the application when the benchmark finishes.
```rust
static mut TICKS: usize = 0;
static mut START_TIME: SystemTime = SystemTime::UNIX_EPOCH;
const SIMULATED_TICKS: usize = 10000;


fn benchmark_system() {
   unsafe {
       if TICKS == 0 {
           START_TIME = SystemTime::now();
       } else if TICKS == SIMULATED_TICKS {
           let duration = SystemTime::elapsed(&START_TIME).unwrap();
           println!("Finished benchmark (took {} ms)", duration.as_millis());
           println!("avg tps {})", (SIMULATED_TICKS as f32) / duration.as_secs_f32());
       }
       TICKS += 1;
   }
}
```

> Which is quite unfortunate, since Bevy is about 25% faster than it was using the common implementation at all body counts, and RECS is 40% slower than it was using floats at all body counts. The slowdown in RECS must be because of using floats, which is very strange since you'd expect more of them to fit into cache/registers. Maybe my CPU only has 64-bit FPUs...
> 
> With the reasoning that we should use all engines in their own ecosystems, I suppose switching back to doubles for RECS would be okay if those perform better in the engine. But maybe you could try running the bench on your machine as well, to see if floats perform better than doubles for you?

On my system, I see around a 30 % improvement for both RECS and Bevy when using floats instead of doubles. It looks very strange on your system. You could maybe try compiling with `-C target-cpu=native` and see if anything changes.

Here is my result with floats:
lines

martinjonsson01:
> This is what I have been using then benchmarking RECS. It is a very hacky solution and I have to manually shutdown the application when the benchmark finishes.
> 
> [...]

I see. That's what I've been trying to avoid in my implementation of these benchmarks, since I'd like them to be "start-and-forget" and entirely automatic. But if that's what's necessary in order to compare to Unity, then we can use it. I'll put that into the PR, along with something similar for Bevy. 

One idea is to create the more rigorous and easy-to-use "manual loop" benchmarks for the engines which allow it (i.e. all but Unity), and use the "engine loop" benchmarks only when it's necessary to compare against Unity. Then when we use this data in the thesis, we only compare alike implementations. That means we could have one huge graph showing all of the engines at the same time (using "engine loop" benches), while also allowing for other graphs that exclude Unity (using "manual loop" benches). 

> On my system, I see around a 30 % improvement for both RECS and Bevy when using floats instead of doubles. It looks very strange on your system. You could maybe try compiling with `-C target-cpu=native` and see if anything changes.
> 
> Here is my result with floats: [...]

Ah, good catch! That was the issue, now the results seem a lot more reasonable:
image

martinjonsson01:
@EdvinNilsson could you take a look at the changes I've made since your last review? It's basically what we've been discussing, I just want to make sure I haven't made any mistakes in the implementation. 

I tried to make my implementation of the "manual" RECS bench as similar to the EnTT one you pushed up earlier today.

martinjonsson01:
Actually, nevermind! I found a nice way of integrating it with Criterion so a lot of the code I pushed becomes unnecessary.

Comparing the "integrated" (i.e. using a benchmark system within the engine) to the ordinary, they're about the same:
image

martinjonsson01:
I realized that it also was strange that we were comparing our entire `Application` to just Bevy's `World`, so I changed the Bevy bench to use a Bevy `App`. This turned out to make a big difference, as we're now quite even:
image

Matte02 (approved):
Good that caught the difference when comparing our whole application and only bevy's world. Still very good to me 👍 

Nice, that we are actually not too far behind in performance compared to bevy.

EdvinNilsson (commented):
I have tested this benchmark and compared it with the old method and the problem is that they do not match up, especially for 1000 bodies where the difference is around 40 %. I have re-runned it multiple times and I cannot understand why there is such a big difference in the middle.


| Body Count | 100      | 500       | 1000     | 10000    |
|------------|----------|-----------|----------|----------|
| Criterion  | 3424,892 | 677,782   | 309,042  | 7,409    |
| Old method | 3662,284 | 691,43695 | 435,8136 | 7,657815 |

martinjonsson01:
> I have tested this benchmark and compared it with the old method and the problem is that they do not match up

Which is "the old method"? If you link me the commit hashes for which versions you've benchmarked, and I'll see if I can reproduce these results on my machine!

EdvinNilsson:
Maybe it something else then. When I enabled tracing on the old method, it was around 40 % slower and performed about the same as Criterion.

martinjonsson01:
> When I enabled tracing on the old method, it was around 40 % slower and performed about the same as Criterion.

Hm, that's oddly fitting. I'll have to double-check.

crates/n-body/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        // Newton's law of universal gravitation.
+        const GRAVITATIONAL_CONSTANT: f32 = 6.67e-11;
+        let acceleration = GRAVITATIONAL_CONSTANT * body_mass / distance_squared;
+
+        to_body.normalize_to(acceleration)
+    };
+
+    let total_acceleration: Vector3<f32> = bodies_query1
```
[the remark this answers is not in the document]
martinjonsson01: Good catch. Clion renamed a bunch of variables when I inlined the functions, unfortunately.

martinjonsson01:
I haven't been able to reproduce the difference. I've created an automated comparison of them both, so if you run the `n_body` bench we can see if you get the same results. 

I get that they're entirely identical (the two lines overlap so much you can't even see the red one): 
image

EdvinNilsson:
With your automated comparison, both implementations perform very similarly, almost identically. When using Criterion with 1000 bodies, it is only measuring a single tick per sample. If I modify the old implementation to return the average over multiple ticks, then I start to see a difference. With 100 ticks, I see around a 20 % better performance and with 1000 ticks, I see around 40 % better performance.

I changed these lines to test this:
```rust
} else if TICKS == SIMULATED_TICKS * 100 { // n_body_old.rs:24

unsafe { DURATION.div(100) } // n_body_old.rs:67
```

martinjonsson01:
> If I modify the old implementation to return the average over multiple ticks, then I start to see a difference. With 100 ticks, I see around a 20 % better performance and with 1000 ticks, I see around 40 % better performance.

I'm still not seeing a difference after applying the 100x averaging:
```
new  [1.7093 ms 1.7320 ms 1.7543 ms]
old   [1.7042 ms 1.7239 ms 1.7440 ms]
```

Even with 1000x averaging:
```
new  [1.7068 ms 1.7321 ms 1.7580 ms]
old   [1.7363 ms 1.7538 ms 1.7716 ms]
```

EdvinNilsson:
I have no idea why there is such a big difference between the implementations on my desktop system. I have tried to debug this with perf, but I could not find anything interesting. 


Here are some graphs from Criterion comparing the new implementation with the old one with `AVERAGE_OVER_TICKS` set to 100.
image

Here is the same benchmark, but running on my laptop instead:
image

Here is more zoomed out version from my desktop:
image

martinjonsson01:
> I have no idea why there is such a big difference between the implementations on my desktop system. I have tried to debug this with perf, but I could not find anything interesting.

Hmm, that's very strange. I was about to suggest we look at it in perf but if even that doesn't show the culprit then I've got no idea. We can discuss this during the meeting today to decide how we should handle it.

EdvinNilsson (commented):
Something is weird with the number from the Criterion csv. I will use the the benchmark results for 1 body and 16 384 bodies as an example.

### For 1 body:
The output from the terminal looks reasonable.
```
n_body/recs/1           time:   [50.220 µs 51.435 µs 52.804 µs]
```

The csv reports 5194981.7 ns (≈ 5.2 ms) on average.
If this number was divided by 100, then it would be reasonable.

### For 16 384 bodies:

Output from terminal:
```
n_body/recs/16384       time:   [355.10 ms 356.77 ms 358.21 ms]
```
The csv reports 356769598.7 ns (≈ 356.7 ms) on average.
Here the result from the terminal and the csv matches and both are reasonable. 

Update:
I think you should divide by `iteration_count` from the csv.

benchmarking/bench_n_body.py
```diff
@@ -0,0 +1,132 @@
...
+import time
+import csv
+import platform
+import subprocess
+
+start_time = time.perf_counter()
+
+print("running Rust benchmarks...")
+os.chdir("..")
+# os.system("cargo bench --bench n_body --features bench-all-engines")
```
EdvinNilsson, line 12:
```suggestion
os.system("cargo bench --bench n_body --features bench-all-engines")
```

EdvinNilsson (approved):
LGTM!

PR #117, opened 2023-04-29
Refactor app execution

martinjonsson01:
Refactors application execution by removing explicit lifetime parameters from `Schedule` and `Executor` traits + removing lifetime parameter from `Task`.

I've (unsuccessfully) tried to do parts of this before, but doing it in only one of the places (schedule, executor, tasks) is not possible, it had to be an engine-wide conversion. I had an epiphany while trying to write a test for #53 that needed to execute a single tick, so I went ahead and implemented this.

This has huge benefits in ergonomics, including:
* The WorkerPool no longer needs to use scoped threads!
  * This should make the implementation of parallel fetching easier to do using the WorkerPool
* Tests can now run ticks manually, without needing overcomplicated shutdown-threads
  * Many tests can be greatly simplified
* Benchmarks can be greatly simplified

The main ways this PR removes lifetime parameters is:
* Cloning `dyn System`s, which is not something usually possible (trait-objects can't return `Self` due to them being unsized). The crate `dyn_clone` made this possible.
* Transferring ownership of all systems into the `Schedule`, rather than have it simply borrow it from `BasicApplication`.
* Wrapping `World` in an `Arc` so each `Task` can keep an owned reference to `World`. This removes the need for the scoped threads in the `WorkerPool` (since even if a `Task` lives longer than the rest of the application, `World` isn't deallocated until that last `Task` is), which should make working with it a lot easier! 

I realize that it looks quite risky performance-wise to spread bunch of `.clone()`s and `Arc::clone`s across the codebase. That didn't turn out to be an issue, though. This is because the clones are not done in the hottest loops of the engine, it's only done once per system per tick and once per system-segment per tick. The following are the (relevant) benchmark results comparing the pre-refactor engine (red) to the post-refactor engine (blue):

### n-body 1 body
n_body_1

### n-body 512 bodies
n_body_512

### n-body 16,384 bodies
n_body_16384

### heavy_compute bench
heavy_compute

### schedule bench
schedule

### DIFF
--- a/crates/ecs/src/lib.rs
+++ b/crates/ecs/src/lib.rs
@@ -41,11 +41,12 @@ use crossbeam::channel::{bounded, Receiver, Sender, TryRecvError};
 use fnv::FnvHashMap;
 use nohash_hasher::{IsEnabled, NoHashHasher};
 use std::any::TypeId;
+use std::cell::{RefCell, RefMut};
 use std::collections::{HashMap, HashSet};
 use std::error::Error;
 use std::fmt::Debug;
 use std::hash::{BuildHasherDefault, Hash};
-use std::sync::{RwLockReadGuard, RwLockWriteGuard};
+use std::sync::{Arc, RwLockReadGuard, RwLockWriteGuard};
 use thiserror::Error;
 
 /// Builds and configures an [`Application`] instance.
@@ -177,8 +178,8 @@ pub trait Application {
 
     /// Starts the application. This function does not return until the shutdown command has
     /// been received.
-    fn run<'systems, E: Executor<'systems>, S: Schedule<'systems>>(
-        &'systems mut self,
+    fn run<E: Executor + 'static, S: Schedule + 'static>(
+        self,
         shutdown_receiver: Receiver<()>,
     ) -> Result<(), Self::Error>;
 }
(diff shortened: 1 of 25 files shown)


crates/ecs/src/lib.rs
```diff
@@ -220,38 +221,89 @@ impl Application for BasicApplication {
...
+impl<E, S> Tickable<E, S> for ApplicationRunner<E, S>
+where
+    S: Schedule,
+    E: Executor,
+{
+    type Error = BasicApplicationError;
+
+    fn tick(&self) -> Result<(), Self::Error> {
+        let mut executor = self.executor.borrow_mut();
+        let mut schedule = self.schedule.borrow_mut();
```
ChristofferPerssons, line 300: Question: 
Does the schedule need to be mutable?

martinjonsson01: Yes, because the executors need the schedule to update its state in order for it to keep track of which systems have finished executing. See here:
https://github.com/martinjonsson01/recs/blob/f406ecdb6fc50b418bf0e3174c19dd2e98c79b14/crates/scheduler/src/executor/mod.rs#L192-L200
on line 200, a call to a function on `schedule` requires `&mut self`.

EdvinNilsson (approved):
LGTM!

Issue #112, opened 2023-04-21
Batch component addition and removal


martinjonsson01:
## Reason
To allow users of the engine to easily and more efficiently create many entities with their components at once, implement a specific API for that.

## Requirements
- [x] There is a function on `Application` called `create_entities` which takes some specification of what components each entity should have 
  - [x] Could be a closure, like Bevy does with their `spawn_batch` function
- [x] `create_entities` creates as many entities as specified, all with the component data given
- [x] Different entities should be able to be initialized with different values for their components 
- [x] This should be done efficiently (i.e. not looping and calling `add_component` repeatedly)
- [x] Matching entities (with same components/in same archetype) with matching commands (what's being added/removed is same on both) should be batched and executed at once

## Dependencies
None

PR #119, opened 2023-05-03
53 system command buffers

### LINKED ISSUE #53: System command buffers

martinjonsson01:
Implements system command buffers, allowing systems to modify the state of the world during app execution.

## Usage

See the tests and doc-tests for usage examples. I've also added an example of using it in our n-body demo. Note that this does not affect the n-body benchmarks.

## Unsafe code

Unfortunately, the only viable way I found to mutate the world in-between ticks is by removing the `Arc` around it and using `unsafe` to pass a `&World` to the system tasks. Please closely review that unsafe code in the precedence graph executor:
https://github.com/martinjonsson01/recs/blob/985aa2c56bef2a3fcfa2851f76613755655b710a/crates/scheduler/src/executor/mod.rs#L210

For context, the options I tried (for way too many hours) before resorting to `unsafe` are:
* Using `Arc<RwLock<World>>`. This is extremely difficult (and bad performance-wise) due to all the places in the engine which require world-references.
* Using lock-free datastructures inside `World` to make it not require `&mut self` (interior mutability). This is problematic since functions like `borrow_component_vecs` need to return references to data within the lock-free data structures and that leaks the guards, leading to the same problem as alternative 1.
* Making `World` contain an `RwLock<InnerWorld>` which then contains all the current `World`-data (interior mutability). This is not viable because of the same reason as alternative 2 (but it leaks `RwLockReadGuard`s from within `InnerWorld` instead).

## Performance

Performance-wise, there (surprisingly) seems to be a great improvement (blue is this PR, red is `master`):

chrome_c3NXp2hQQN

This might be because I've switched our `RwLock`s from `std::sync::RwLock` to `parking_lot::RwLock` which is supposed to be a faster version of the same thing.

Closes #53


crates/ecs/src/systems/command_buffers.rs
```diff
@@ -0,0 +1,715 @@
...
+            "three new entities should have been created"
+        );
+        assert_eq!(
+            a_values.len(),
+            3,
+            "3 entities with components A, B and C should have been created"
+        );
+        assert_eq!(a_values.len(), b_values.len());
+        assert_eq!(b_values.len(), c_values.len());
+    }
```
EdvinNilsson, line 714: Found a problem that causes a panic if you add and remove entities repeatedly. Here is a simple test for that:
```suggestion
    }
    
    #[test]
    fn creating_and_removing_entities_multiple_times_does_not_cause_panic() {
        let removing_creation_system = |entity: Entity, commands: Commands| {
            commands.remove(entity);
            let creation = EntityCreation::default().with_component(C(1.0));
            commands.create(creation);
        };

        let (app, _, _, _) = set_up_app_with_systems_and_entities([removing_creation_system]);

        let mut runner = app.into_tickable::<Sequential, Unordered>().unwrap();

        for _ in 0..10 {
            runner.tick().unwrap();
            runner.playback_commands().unwrap();
        }
    }
```

martinjonsson01: Thanks, I'll fix that!

EdvinNilsson (approved):
LGTM!

I don't know if something is wrong with your performance graphs, but I do not see this big performance increase with this PR. Instead, I see the opposite (blue is this PR).

Screenshot_20230503_151706

crates/ecs/src/lib.rs
```diff
@@ -573,6 +601,125 @@ pub struct World {
...
+        let entities: Vec<_> = entities.into_iter().map(Result::unwrap).collect();
+        let failures: Vec<_> = failures.into_iter().map(Result::unwrap_err).collect();
+        if failures.is_empty() {
+            Ok(entities)
+        } else {
+            Err(WorldError::EntityCreation(entities, failures))
+        }
+    }
+
+    fn create_entity(&mut self, creation: EntityCreation) -> WorldResult<Entity> {
```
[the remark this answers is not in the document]
martinjonsson01: Yeah, that's fair enough. Although I should note that you usually use error-variants along with their containing error type so you could still distinguish `EntityCreation` from `WorldError::EntityCreation`. 

No matter, though, as I've made some great API-usability improvements in an upcoming PR for #112 which I'll be publishing in a moment. `EntityCreation` is removed in that PR.

martinjonsson01:
> I do not see this big performance increase with this PR. Instead, I see the opposite

Hm, that does seem to correspond with what the GitHub-runner is seeing as well.

Maybe my computer is haunted... 

Oh well, that's much more reasonable anyways seeing as more functionality has been added to the addition/removal-machinery.

crates/ecs/src/systems/command_buffers.rs
```diff
@@ -0,0 +1,742 @@
...
+        runner.playback_commands().unwrap();
+
+        let remaining_entities = remaining_entities_in_world(&runner.world);
+        assert!(
+            remaining_entities.is_empty(),
+            "all entities should be removed, but these remain: {:?}",
+            remaining_entities
+        )
+    }
+
```
JacobBredin, line 527: I found an issue where if you have **one system adding components** and **one removing the same entities**, it results in an unexpected crash, caused by `.expect("Entity should have a component index tied to it since it is already established to be existing within the archetype")` on line 477 in archetypes.rs.

JacobBredin (approved):
LGTM! 

I have looked through your code and I like implementation you have done. I found one issue with the commands. I don't think you need to fix it, but know that it is there is probably good when implementing the rain simulation.

crates/ecs/src/archetypes.rs
```diff
@@ -456,15 +533,36 @@ impl World {
...
+        entity: Entity,
+        components: impl IntoIterator<Item = Box<dyn AnyComponent>>,
+    ) -> WorldResult<()> {
         let source_archetype_index = *self
             .entity_to_archetype_index
             .get(&entity)
             .ok_or(WorldError::EntityDoesNotExist(entity))?;
 
-        let add = ArchetypeMutation::Addition(entity);
+        let components: Vec<_> = components.into_iter().collect();
```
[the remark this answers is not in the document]
martinjonsson01: The collect is needed so I can create multiple iterators from it. The iterator in the function signature is an owning iterator, meaning it's `Item` is an owned `Box<dyn AnyComponent>`. That's necessary because the function needs to take ownership of the new components to store them in the archetype. I considered making it `Clone` but then every component within it would also have to be `Clone`, which seems a bit too strict. 

If we require all components to be `Clone`, then it would probably be possible to avoid this allocation by cloning the iterator. What do you think?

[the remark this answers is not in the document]
JacobBredin: Ok fair enough 👍

crates/ecs/src/systems/command_buffers.rs
JacobBredin: If you do the same thing but with `commands.remove_component` it works correctly, so it seems reasonable that both cases should be allowed.

JacobBredin: All seems to work! Close this and MERGE !

PR #120, opened 2023-05-03
112 batch component addition and removal

### LINKED ISSUE #112: Batch component addition and removal

martinjonsson01:
Implements a much more concise entity-creation API.

Although this is medium priority, I noticed that I had already implemented the majority of this issue in #119, so I went ahead and finished it since it was only marginally more work. 

### API Changes
The new API includes the following additions on `Application`:
```rust
/// Spawns an entity with no components.
fn create_empty_entity(&mut self) -> Result<Entity, Self::Error>;

/// Spawns a new entity with the given components.
fn create_entity(
    &mut self,
    components: impl IntoBoxedComponentIter,
) -> Result<Entity, Self::Error>;

/// Spawns multiple new entities with the given components.
fn create_entities(
    &mut self,
    creations: impl IntoIterator<Item = impl IntoBoxedComponentIter>,
) -> Result<Vec<Entity>, Self::Error>;

/// Spawns multiple new entities containing the same set of components.
///
/// The provided closure is called once for each created entity in order to
/// uniquely instantiate their values.
fn create_entities_with<RangeIndex, Components>(
    &mut self,
    number_of_entities: RangeIndex,
    components_creator: impl FnMut(RangeIndex) -> Components,
) -> Result<Vec<Entity>, Self::Error>
where
    RangeIndex: Num + PartialOrd + Step,
    Components: IntoBoxedComponentIter;
```

I've updated all usages, so the changes speak for themselves:
chrome_Bc7E2mQ84j

chrome_DEPEsFvpLx

chrome_CkpOGX1s5b

### Performance changes
According to [the GitHub-runner](https://github.com/martinjonsson01/recs/pull/120#issuecomment-1533151237) there's a **76.92%** improvement in the `simple_insert` benchmark!

Closes #112


ChristofferPerssons (approved):
LGTM!

EdvinNilsson (approved):
LGTM!

PR #121, opened 2023-05-06
Replace rayon with workerpool

martinjonsson01:
Replaces the Rayon-based parallel parameter fetching with WorkerPool-based system segmentation.

I'm not very happy with how it turned out in the end, needing to continue our use of `unsafe`. It can definitely be done much more cleanly and better, but I couldn't find that solution at the moment.

It's functional at least, which is what's important for the benchmarks.


EdvinNilsson (approved):
Great that you got this working! Although this is not the most elegant solution, it is good enough. I can fix the last things with parallel fetching.

martinjonsson01:
Since this isn't a full-fledged issue, do you think it's fine if we merge it immediately now? I'm thinking that since this is only a small part of the complete issue of parallel parameter fetching, it doesn't need another review (to expedite things a bit).

EdvinNilsson:
You can merge it to `89-parallel-parameter-fetching`.

